PR #62, opened 2019-11-07
Allow create to use env overrides

nckturner:
*Description of changes:*
- Create cluster and create config allows env var overrides.  Fixes: https://github.com/aws/aws-k8s-tester/issues/63
- Remove extra ValidateAndSetDefaults() which gets called twice
  in the current create cluster path.


gyuho:
lgtm thx!

Issue #66, opened 2019-12-23
attempt to chmod `aws` binary even if it's already world-executable


jaypipes:
ref: https://github.com/aws/amazon-vpc-cni-k8s/issues/770

We want to use an updated aws-k8s-tester binary (UpdateFromEnv was added in 0.5.1) but unfortunately v0.5.1 fails for me because it attempts to `chmod` the aws CLI:

```
failed to create EKS deployer error doing chmod on "/usr/local/bin/aws": chmod /usr/local/bin/aws: operation not permitted
```

Note that I have the aws binary already installed and is world-executable:

```
jaypipes@thelio:~/go/src/github.com/aws/amazon-vpc-cni-k8s$ ls -l `which aws`
-rwxr-xr-x 1 root root 815 Jun 28 10:25 /usr/local/bin/aws
```

so I'm not sure why aws-k8s-tester is trying to chmod anything.

PR #67, opened 2019-12-23
Only attempt chmod of `aws` CLI if not executable

jaypipes:
Adds a file stat check to ensure we only attempt to do a chmod on the
`aws` CLI executable if it does not already have execute permissions.

Closes Issue #66

### DIFF
--- a/pkg/fileutil/fileutil.go
+++ b/pkg/fileutil/fileutil.go
@@ -7,6 +7,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"syscall"
 	"time"
 )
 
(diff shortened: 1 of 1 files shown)


pkg/fileutil/fileutil.go
```diff
@@ -79,6 +80,10 @@ func EnsureExecutable(p string) error {
 	if err != nil {
 		return fmt.Errorf("error doing stat on %q: %v", p, err)
 	}
+	m := s.Mode()
+	if m&(syscall.S_IXOTH|syscall.S_IXGRP|syscall.S_IXUSR) != 0 {
+		return nil
+	}
 	if err := os.Chmod(p, s.Mode()|0111); err != nil {
```
mogren: Nice use of `syscall.S_IXOTH|syscall.S_IXGRP|syscall.S_IXUSR`! Should we do 
```
os.Chmod(p, s.Mode()|syscall.S_IXOTH|syscall.S_IXGRP|syscall.S_IXUSR)
``` 
as well instead of using `0111`?

leakingtapan: Curious what's the harm if chmod an executable again? From the error message below from the linked issue, it seems a permission issue to me. Could it because the `aws` is owned by `root` while `aws-k8s-tester` is running as some other user?

```
failed to create EKS deployer error doing chmod on "/usr/local/bin/aws": chmod /usr/local/bin/aws: operation not permitted
```

jaypipes: Yes, that's precisely what is the case. /usr/local/bin/aws is owned by root for some reason. But the point of this patch is that EnsureExecutable is merely looking to ensure the aws CLI tool is executable, nothing more, and chmod'ing a file that is owned by root isn't necessary to check for execute permissions.

Issue #70, opened 2020-01-02
unable to detach ENI that was created by aws-k8s-tester due to permissions issue


jaypipes:
When attempting to delete a cluster that was successfully created using `aws-k8s-tester eks create cluster`, I am getting "failed to detach ENI: AuthFailure: You do not have permission to access the specified resource" errors.

Here is the log output of the time in question, showing deleting the VPC CloudFormation stack failing, force-deleting subnets failing because of dependency violations, and then failure to detach ENIs that subnets were dependent on due to permissions issues:

```
{"level":"info","ts":"2020-01-02T09:35:17.542-0500","caller":"eks/vpc.go:367","msg":"deleting VPC CFN stack","vpc-cfn-stack-id":"arn:aws:cloudformation:us-west-2:750630568209:stack/eks-2020010213-pdx-us-west-2-belcb-vpc/04d0d020-2d62-11ea-a307-02ca0801967a"}
{"level":"info","ts":"2020-01-02T09:35:18.222-0500","caller":"cloudformation/cloudformation.go:40","msg":"polling stack","stack-id":"arn:aws:cloudformation:us-west-2:750630568209:stack/eks-2020010213-pdx-us-west-2-belcb-vpc/04d0d020-2d62-11ea-a307-02ca0801967a","want":"DELETE_COMPLETE"}
{"level":"info","ts":"2020-01-02T09:35:38.623-0500","caller":"cloudformation/cloudformation.go:112","msg":"sleeping","stack-id":"arn:aws:cloudformation:us-west-2:750630568209:stack/eks-2020010213-pdx-us-west-2-belcb-vpc/04d0d020-2d62-11ea-a307-02ca0801967a","initial-wait":"1m30s"}
{"level":"info","ts":"2020-01-02T09:37:08.624-0500","caller":"cloudformation/cloudformation.go:141","msg":"polling","stack-id":"arn:aws:cloudformation:us-west-2:750630568209:stack/eks-2020010213-pdx-us-west-2-belcb-vpc/04d0d020-2d62-11ea-a307-02ca0801967a","stack-name":"eks-2020010213-pdx-us-west-2-belcb-vpc","current":"DELETE_IN_PROGRESS","want":"DELETE_COMPLETE","reason":"","request-started":"1 minute ago"}
[...]

PR #71, opened 2020-01-02
wait before attempting to delete k8s namespace

jaypipes:
Wait for before trying to delete the k8s namespace on cluster down and
when MNG and the ALB/NLB testers have been run.

Because we asked Kubernetes to delete the NLB hello world and ALB2048
Deployment/Service above, and both of these interact with the underlying
Kubernetes AWS cloud provider to clean up the cloud load balancer
backing the Service of type LoadBalancer. The calls to delete the
Service return immediately (successfully) but the cloud load balancer
resources may not have been deleted yet, including the ENIs that were
associated with the cloud load balancer. When, later, aws-k8s-tester
tries deleting the VPC associated with the test cluster, it will run
into permissions issues because the IAM role that created the ENIs
associated with the ENIs in subnets associated with the cloud load
balancers will no longer exist.

Issue #70 

### DIFF
--- a/eks/eks.go
+++ b/eks/eks.go
@@ -486,6 +486,29 @@ func (ts *Tester) down() (err error) {
 		}
 	}
 
+	// NOTE(jaypipes): Wait for a bit here because we asked Kubernetes to
+	// delete the NLB hello world and ALB2048 Deployment/Service above, and
+	// both of these interact with the underlying Kubernetes AWS cloud provider
+	// to clean up the cloud load balancer backing the Service of type
+	// LoadBalancer. The calls to delete the Service return immediately
+	// (successfully) but the cloud load balancer resources may not have been
+	// deleted yet, including the ENIs that were associated with the cloud load
+	// balancer. When, later, aws-k8s-tester tries deleting the VPC associated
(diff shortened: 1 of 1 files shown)


eks/eks.go
```diff
@@ -486,6 +486,29 @@ func (ts *Tester) down() (err error) {
...
+	// IAM role that created the ENIs associated with the ENIs in subnets
+	// associated with the cloud load balancers will no longer exist.
+	//
+	// https://github.com/aws/aws-k8s-tester/issues/70
+	// https://github.com/kubernetes/kubernetes/issues/53451
+	// https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/20190423-service-lb-finalizer.md
+	if ts.cfg.Parameters.ManagedNodeGroupCreate &&
+		(ts.cfg.AddOnALB2048.Enable && ts.alb2048Tester != nil ||
+			ts.cfg.AddOnNLBHelloWorld.Enable && ts.nlbHelloWorldTester != nil) {
+		waitDur := 30 * time.Second
```
mogren: Is 30s just an educated guess, or do we have any numbers on how long this usually takes, or what the worst case time is?

PR #72, opened 2020-01-03
ensure subtester jobs are torn down properly

jaypipes:
There was a problem with ALB/NLB resources not being cleaned up
when an EKS cluster was deleted by aws-k8s-tester. It turned out that
the subtester objects that create/delete resources like the ALB2048
Deployment were not being constructed properly if the
`aws-k8s-tester delete cluster` command was run. Previously, these
subtester objects were only created in `Tester.up()` which is not called
when `Tester.down()` is called.

This commit moves construction of these subtester objects into the
construction of the primary Tester object, which alleviates the issue
with non-initialized subtesters in `Tester.down()`. In doing so, I've
changed the construction of the subtester objects to take a pointer
to the `Tester` instance and grab their K8sClientSets from that
instance.

Issue #70 

### DIFF
--- a/eks/alb/alb.go
+++ b/eks/alb/alb.go
@@ -35,7 +35,11 @@ type Config struct {
 	Sig               chan os.Signal
 	EKSConfig         *eksconfig.Config
 	CloudFormationAPI cloudformationiface.CloudFormationAPI
-	K8SClient         *clientset.Clientset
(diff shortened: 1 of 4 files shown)


jaypipes:
I've tested this functionally with CNI plugin and everything is now running smoothly, with no orphaned resources or failure to delete/detach resources.

PR #87, opened 2020-05-28
*: add cluster loader local/remote

gyuho:
WIP


gyuho:
Requires https://github.com/kubernetes/perf-tests/pull/1295.

PR #97, opened 2020-06-12
cluster autoscaler test

hakuna-matatah:
*Description of changes:*
Add cluster autoscaler test as part of node group test suite.


eks/ng/nodes.go
```diff
@@ -243,6 +243,10 @@ Resources:
         PropagateAtLaunch: true
       - Key: !Sub kubernetes.io/cluster/${ClusterName}
         Value: owned
+      - Key: !Sub k8s.io/cluster-autoscaler/${ClusterName}
```
gyuho, line 246: Can we make it as template? e.g. https://github.com/aws/aws-k8s-tester/blob/master/eks/ng/nodes.go#L211

eks/ng/role.go
```diff
@@ -195,7 +195,17 @@ Resources:
             - route53:ChangeResourceRecordSets
             - route53:DeleteHealthCheck
             Resource: "*"
-
+          - Effect: Allow
```
gyuho, line 198: Also this should be conditional, only added when cluster autoscaler is enabled.

gyuho:
Does CA require metrics-server? Any other dependency? If so, we need err when metrics server is not installed (right now installed via `kubernetes-dashboard` add-on)

gyuho:
And if this has to be tied to node group, maybe we can add this as a nested add-on such as mng version upgrade (e.g. https://github.com/aws/aws-k8s-tester/tree/master/eks/mng/version-upgrade).

Also, can this be integrated with MNG?

gyuho:
Also, what would happen if we run cluster autoscaler with MNG? Would it conflict with MNG scale config API call? Maybe for the first iteration, we can focus on NG autoscaler.

eks/ng/clusterautoscaler.go
```diff
@@ -0,0 +1,454 @@
+package ng
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"github.com/aws/aws-k8s-tester/pkg/fileutil"
```
gyuho, line 7: We should move this to `eks/ng/autoscaler`

gyuho: And automatically installed when `NG` has `"autoscaler":{"enable":true...` field

eks/cluster-autoscaler/cluster_autoscaler.go
```diff
@@ -0,0 +1,19 @@
+package cluster_autoscaler
```
gyuho, line 1: Let's move it to `eks/ng/autoscaler`?

eks/cluster-autoscaler/cluster_autoscaler.go
```diff
@@ -0,0 +1,19 @@
...
+
+import (
+	"github.com/aws/aws-k8s-tester/eksconfig"
+	k8s_client "github.com/aws/aws-k8s-tester/pkg/k8s-client"
+	"github.com/aws/aws-sdk-go/service/cloudformation/cloudformationiface"
+	"go.uber.org/zap"
```
gyuho, line 7: Let's place standard packages (e.g. `"os"` at top, and 3rd party packages bottom, and run `./scripts/fmt.sh`)

eks/ng/clusterautoscaler.go
```diff
@@ -0,0 +1,454 @@
...
+  - apiGroups: ["batch", "extensions"]
+    resources: ["jobs"]
+    verbs: ["get", "list", "watch", "patch"]
+  - apiGroups: ["coordination.k8s.io"]
+    resources: ["leases"]
+    verbs: ["create"]
+  - apiGroups: ["coordination.k8s.io"]
+    resourceNames: ["cluster-autoscaler"]
+    resources: ["leases"]
+    verbs: ["get", "update"]
```
gyuho, line 82: Can add empty lines between here to make it more readable

eks/ng/clusterautoscaler.go
```diff
@@ -0,0 +1,454 @@
...
+              cpu: 100m
+              memory: 300Mi
+          command:
+            - ./cluster-autoscaler
+            - --v=4
+            - --stderrthreshold=info
+            - --cloud-provider=aws
+            - --skip-nodes-with-local-storage=false
+            - --expander=least-waste
+            - --node-group-auto-discovery=asg:tag=k8s.io/cluster-autoscaler/enabled,k8s.io/cluster-autoscaler/<YOUR CLUSTER NAME>
```
gyuho, line 175: `<YOUR CLUSTER NAME>` should be parameter in template

Issue #91, opened 2020-06-09
Support EKS node group update add-on "AddOnMNG/NGUpdate"


gyuho:
Scale up existing ng/mng from x to y.

gyuho:
Done via https://github.com/aws/aws-k8s-tester/pull/106.

PR #99, opened 2020-06-15
add cluster autoscaler addon

hakuna-matatah:
*Description of changes:*
add cluster autoscaler addon.
Clusteraddon tests can be enabled like any other addon by setting env variable
It's disabled by default.

### DIFF
--- a/eks/ng/autoscaler/cluster-autoscaler.go
+++ b/eks/ng/autoscaler/cluster-autoscaler.go
@@ -1,0 +1,398 @@
+package autoscaler
+
+import (
+	"bytes"
+	"context"
+	"errors"
+	"fmt"
+	"reflect"
+	"strings"
+	"text/template"
+	"time"
+
+	"github.com/aws/aws-k8s-tester/eksconfig"
+	"github.com/aws/aws-k8s-tester/pkg/fileutil"
+	k8s_client "github.com/aws/aws-k8s-tester/pkg/k8s-client"
+	"github.com/aws/aws-sdk-go/service/autoscaling/autoscalingiface"
+	"github.com/aws/aws-sdk-go/service/ec2/ec2iface"
+	"github.com/aws/aws-sdk-go/service/eks/eksiface"
+	"go.uber.org/zap"
+	appsv1 "k8s.io/api/apps/v1"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/exec"
+)
+
+// reference - https://raw.githubusercontent.com/kubernetes/autoscaler/master/cluster-autoscaler/cloudprovider/aws/examples/cluster-autoscaler-autodiscover.yaml
+const ClusterAutoscalerYAML = `
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  labels:
+    k8s-addon: cluster-autoscaler.addons.k8s.io
+    k8s-app: cluster-autoscaler
+  name: cluster-autoscaler
+  namespace: kube-system
+
+---
+
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: cluster-autoscaler
+  labels:
+    k8s-addon: cluster-autoscaler.addons.k8s.io
+    k8s-app: cluster-autoscaler
+rules:
+  - apiGroups: [""]
+    resources: ["events", "endpoints"]
+    verbs: ["create", "patch"]
+  - apiGroups: [""]
+    resources: ["pods/eviction"]
+    verbs: ["create"]
+  - apiGroups: [""]
+    resources: ["pods/status"]
+    verbs: ["update"]
+  - apiGroups: [""]
+    resources: ["endpoints"]
+    resourceNames: ["cluster-autoscaler"]
+    verbs: ["get", "update"]
+  - apiGroups: [""]
+    resources: ["nodes"]
+    verbs: ["watch", "list", "get", "update"]
+  - apiGroups: [""]
+    resources:
+      - "pods"
+      - "services"
+      - "replicationcontrollers"
+      - "persistentvolumeclaims"
+      - "persistentvolumes"
+    verbs: ["watch", "list", "get"]
+  - apiGroups: ["extensions"]
+    resources: ["replicasets", "daemonsets"]
+    verbs: ["watch", "list", "get"]
+  - apiGroups: ["policy"]
+    resources: ["poddisruptionbudgets"]
+    verbs: ["watch", "list"]
+  - apiGroups: ["apps"]
+    resources: ["statefulsets", "replicasets", "daemonsets"]
+    verbs: ["watch", "list", "get"]
+  - apiGroups: ["storage.k8s.io"]
+    resources: ["storageclasses", "csinodes"]
+    verbs: ["watch", "list", "get"]
+  - apiGroups: ["batch", "extensions"]
+    resources: ["jobs"]
+    verbs: ["get", "list", "watch", "patch"]
+  - apiGroups: ["coordination.k8s.io"]
+    resources: ["leases"]
+    verbs: ["create"]
+  - apiGroups: ["coordination.k8s.io"]
+    resourceNames: ["cluster-autoscaler"]
+    resources: ["leases"]
+    verbs: ["get", "update"]
+
+---
+
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: cluster-autoscaler
+  namespace: kube-system
+  labels:
+    k8s-addon: cluster-autoscaler.addons.k8s.io
+    k8s-app: cluster-autoscaler
+rules:
+  - apiGroups: [""]
+    resources: ["configmaps"]
+    verbs: ["create","list","watch"]
+  - apiGroups: [""]
+    resources: ["configmaps"]
+    resourceNames: ["cluster-autoscaler-status", "cluster-autoscaler-priority-expander"]
+    verbs: ["delete", "get", "update", "watch"]
+
+---
+
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: cluster-autoscaler
+  labels:
+    k8s-addon: cluster-autoscaler.addons.k8s.io
+    k8s-app: cluster-autoscaler
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: cluster-autoscaler
+subjects:
+  - kind: ServiceAccount
+    name: cluster-autoscaler
+    namespace: kube-system
+
+---
+
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: cluster-autoscaler
+  namespace: kube-system
+  labels:
+    k8s-addon: cluster-autoscaler.addons.k8s.io
+    k8s-app: cluster-autoscaler
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: cluster-autoscaler
+subjects:
+  - kind: ServiceAccount
+    name: cluster-autoscaler
+    namespace: kube-system
+
+---
+
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: cluster-autoscaler
+  namespace: kube-system
+  labels:
+    app: cluster-autoscaler
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: cluster-autoscaler
+  template:
+    metadata:
+      labels:
+        app: cluster-autoscaler
+      annotations:
+        prometheus.io/scrape: 'true'
+        prometheus.io/port: '8085'
+        cluster-autoscaler.kubernetes.io/safe-to-evict: 'false'
+    spec:
+      serviceAccountName: cluster-autoscaler
+      containers:
+{{ if ne .ImageURI "" }}{{.ImageURI}}{{ end }}
+          name: cluster-autoscaler
+          resources:
+            limits:
+              cpu: 100m
+              memory: 300Mi
+            requests:
+              cpu: 100m
+              memory: 300Mi
+          command:
+            - ./cluster-autoscaler
+            - --v=4
+            - --stderrthreshold=info
+            - --cloud-provider=aws
+            - --skip-nodes-with-local-storage=false
+            - --expander=least-waste
+{{ if ne .NodeGroupAutoDiscovery "" }}{{.NodeGroupAutoDiscovery}}{{ end }}
+            - --balance-similar-node-groups
+            - --skip-nodes-with-system-pods=false
+          volumeMounts:
+            - name: ssl-certs
+              mountPath: /etc/ssl/certs/ca-certificates.crt
+              readOnly: true
+          imagePullPolicy: "Always"
+      volumes:
+        - name: ssl-certs
+          hostPath:
+            path: "/etc/ssl/certs/ca-bundle.crt"
+`
+
+var images = map[string]string{
+	"1.17": `        - image: us.gcr.io/k8s-artifacts-prod/autoscaling/cluster-autoscaler:v1.17.2`,
+	"1.16": `        - image: us.gcr.io/k8s-artifacts-prod/autoscaling/cluster-autoscaler:v1.16.5`,
+}
+
+const (
+	nodeGroupAuotDiscoveryData      = `            - --node-group-auto-discovery=asg:tag=k8s.io/cluster-autoscaler/enabled,k8s.io/cluster-autoscaler/`
+	clusterAutoScalerDeploymentName = "cluster-autoscaler"
+)
+
+type caSpecData struct {
+	ImageURI               string
+	NodeGroupAutoDiscovery string
+}
+
+// Config defines version upgrade configuration.
+type Config struct {
+	Logger    *zap.Logger
+	Stopc     chan struct{}
+	EKSConfig *eksconfig.Config
+	K8SClient k8s_client.EKS
+	EC2API    ec2iface.EC2API
+	ASGAPI    autoscalingiface.AutoScalingAPI
+	EKSAPI    eksiface.EKSAPI
+}
+
+// ClusterAutoScaler defines cluster autoscaler operation.
+type ClusterAutoscaler interface {
+	Create() error
+}
+
+// New creates a new cluster autoscaler.
+func New(cfg Config) ClusterAutoscaler {
+	cfg.Logger.Info("creating tester", zap.String("tester", reflect.TypeOf(tester{}).PkgPath()))
+	return &tester{cfg: cfg}
+}
+
+type tester struct {
+	cfg Config
+}
+
+func (ts *tester) Create() error {
+	needInstall := false
+	for _, cur := range ts.cfg.EKSConfig.AddOnNodeGroups.ASGs {
+		if cur.ClusterAutoScaler.Enable {
+			needInstall = true
+			break
+		}
+	}
+	if !needInstall {
+		ts.cfg.Logger.Info("no NG enables CA; skipping")
+		return nil
+	}
+	return ts.installCA()
+}
+
+func (ts *tester) installCA() error {
+	ts.cfg.Logger.Info("creating CA using kubectl", zap.String("name", ts.cfg.EKSConfig.Name))
+	var ok bool
+	var caData = caSpecData{}
+	caData.ImageURI, ok = images[ts.cfg.EKSConfig.Parameters.Version]
+	if !ok {
+		return fmt.Errorf("no CA found for %q", ts.cfg.EKSConfig.Parameters.Version)
+	}
+	caData.NodeGroupAutoDiscovery = nodeGroupAuotDiscoveryData + ts.cfg.EKSConfig.Name
+	tpl := template.Must(template.New("TemplateCA").Parse(ClusterAutoscalerYAML))
+	buf := bytes.NewBuffer(nil)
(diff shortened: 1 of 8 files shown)


eks/ng/role.go
```diff
@@ -205,8 +208,30 @@ Outputs:
...
+            - ec2:DescribeLaunchTemplateVersions
+            Resource: "*"
+`
+
 func (ts *tester) createRole() error {
 	if !ts.cfg.EKSConfig.AddOnNodeGroups.RoleCreate {
+		// if role isn't created and autoscaler is enabled then return error
+		// because we cannot assume ASG policy is attached to node iam role
+		// this check is also enabled when calling createCA() in nodes.go which comes after createRole() func call
+		if ts.cfg.EKSConfig.IsEnabledAddOnClusterAutoscaler() {
```
gyuho, line 232: Remove `ts.cfg.EKSConfig.IsEnabledAddOnClusterAutoscaler` since we are adding autoscaler inside node group `struct`. You can iterate `NGs` and store a boolean variable to indicate whether to add ASG policy or not.

eksconfig/add-on-cluster-autoscaler.go
```diff
@@ -0,0 +1,54 @@
+package eksconfig
```
gyuho, line 1: This should be inside `AddOnNodeGroups`

eksconfig/config.go
```diff
@@ -303,9 +303,12 @@ type Config struct {
 	// AddOnClusterVersionUpgrade defines parameters
 	// for EKS cluster version upgrade add-on.
 	AddOnClusterVersionUpgrade *AddOnClusterVersionUpgrade `json:"add-on-cluster-version-upgrade,omitempty"`
+	// AddOnClusterAutoscaler defines parameters for EKS Cluster
+	// add on clusterautoscaler
+	AddOnClusterAutoscaler *AddOnClusterAutoscaler `json:"add_on_cluster_autoscaler,omitempty"`
```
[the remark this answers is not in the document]
gyuho: That way, we can configure cluster auto scaler "per" node group. See how we implement `version-upgrade` for MNG

gyuho:
You don't need create a new PR. You can push the changes to the same branch.

eksconfig/config.go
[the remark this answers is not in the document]
gyuho: Similar to https://github.com/aws/aws-k8s-tester/issues/91.

eks/ng/role.go
```diff
@@ -196,6 +198,7 @@ Resources:
             - route53:ChangeResourceRecordSets
             - route53:DeleteHealthCheck
             Resource: "*"
+{{ if ne .UserData "" }}{{.UserData}}{{ end }}
```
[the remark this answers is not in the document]
hakuna-matatah: I can rename but UserData is not being used in this file at all ( here there is no EC2 bootstrap data for role)

eksconfig/config.go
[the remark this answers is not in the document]
hakuna-matatah: Ah! per node group. More flexibility than enabling CA for every node group. I like it. I will do that.

eks/ng/role.go
hakuna-matatah: Sure, that makes more sense ( as we are planning to move it inside the NG)

eksconfig/config.go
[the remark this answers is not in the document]
hakuna-matatah: > ```
> type AddOnNodeGroups struct {
>     ...
>     AutoScaler *NGAutoScaler `json:"auto-scaler"`
>     ....
> ```
> 
> ```
> type NGAutoScaler struct {
>     Enable bool `json:"enable"`
>     ...
> ```

Actually it makes sense to add to ASG Struct than AddonNodeGroupStruct 
```
type NGClusterAutoScaler struct {
	Enable bool `json:"enable"`
	// Created is true when the resource has been created.
	// Used for delete operations.
	Created         bool               `json:"created" read-only:"true"`
	TimeFrameCreate timeutil.TimeFrame `json:"time-frame-create" read-only:"true"`
}
```
```
type ASG struct {
	...
	ClusterAutoScaler *NGClusterAutoScaler `json:"cluster_auto_scaler,omitempty"`
...
}
```

eksconfig/add-on-cluster-autoscaler.go
hakuna-matatah: > This should be inside `AddOnNodeGroups`

This doesn't make sense anymore as a separate file if we are planning to take the route of embedding inside NodeGroup/ASG.

eks/ng/cluster-autoscaler/cluster-autoscaler.go
```diff
@@ -0,0 +1,396 @@
...
+
+type tester struct {
+	cfg Config
+}
+
+func (ts *tester) CreateCA() error {
+	return ts.createCA()
+}
+
+func (ts *tester) createCA() error {
```
gyuho, line 250: Can we iterate all `ts.cfg.EKSConfig.AddOnNodeGroups.ASGs` and check if we need install CA?

```
needInstall := false
for ... range ts.cfg.EKSConfig.AddOnNodeGroups.ASGs {
   if cur.Autoscaler.Enable { needInstall = true; break; }
   ...

if !needInstall { lg.Info("no NG enables CA; skipping"); return nil; }
```

eksconfig/README.md
```diff
@@ -36,6 +36,7 @@ AWS_K8S_TESTER_EKS_ADD_ON_STRESSER_REMOTE_ENABLE=true \
 AWS_K8S_TESTER_EKS_ADD_ON_CLUSTER_LOADER_LOCAL_ENABLE=true \
 AWS_K8S_TESTER_EKS_ADD_ON_CLUSTER_LOADER_REMOTE_ENABLE=true \
 AWS_K8S_TESTER_EKS_ADD_ON_CLUSTER_VERSION_UPGRADE_ENABLE=true \
+AWS_K8S_TESTER_ADD_ON_CLUSTER_AUTOSCALER_ENABLE=true \
```
gyuho, line 39: We do not want to cluster-wide autoscaler enable. This should be scoped down to individual node groups. I think your code already does that. Can we remove this?

eksconfig/add-on-node-groups.go
```diff
@@ -68,6 +76,8 @@ type ASG struct {
 	// TODO: handle conflicting flag '--cloud-provider aws'
 	// ref. https://github.com/kubernetes/kubernetes/issues/64659
 	KubeletExtraArgs string `json:"kubelet-extra-args"`
+	// clusterautoscaler per nodegroup
+	ClusterAutoScaler *NGClusterAutoScaler `json:"cluster_auto_scaler,omitempty"`
```
gyuho, line 80: To be consistent with other JSON tags, let's just name it `json:"cluster-autoscaler,omitempty"`

eksconfig/env_test.go
```diff
@@ -141,7 +141,7 @@ func TestEnv(t *testing.T) {
...
+	os.Setenv("AWS_K8S_TESTER_EKS_ADD_ON_NODE_GROUPS_ASGS", `{"ng-test-name-cpu":{"name":"ng-test-name-cpu","remote-access-user-name":"ec2-user","ami-type":"AL2_x86_64","image-id-ssm-parameter":"/aws/service/eks/optimized-ami/1.30/amazon-linux-2/recommended/image_id","asg-min-size":17,"kubelet-extra-args":"bbb qq", "cluster_auto_scaler" : {"enable" : false}, asg-max-size":99,"asg-desired-capacity":77,"instance-types":["type-cpu-2"],"volume-size":40},"ng-test-name-gpu":{"name":"ng-test-name-gpu","remote-access-user-name":"ec2-user","ami-type":"AL2_x86_64_GPU","asg-min-size":30,"asg-max-size":35,"asg-desired-capacity":34,"instance-types":["type-gpu-2"],"image-id":"my-gpu-ami","volume-size":500, "cluster_auto_scaler": {"enable":false},"kubelet-extra-args":"aaa aa"}}`)
```
gyuho, line 144: This should be `cluster-autoscaler`.

eksconfig/env_test.go
```diff
@@ -1352,7 +1354,7 @@ func TestEnvAddOnNodeGroupsGetRef(t *testing.T) {
 
 	os.Setenv("AWS_K8S_TESTER_EKS_ADD_ON_NODE_GROUPS_ENABLE", `true`)
 	defer os.Unsetenv("AWS_K8S_TESTER_EKS_ADD_ON_NODE_GROUPS_ENABLE")
-	os.Setenv("AWS_K8S_TESTER_EKS_ADD_ON_NODE_GROUPS_ASGS", `{"GetRef.Name-ng-for-cni":{"name":"GetRef.Name-ng-for-cni","remote-access-user-name":"ec2-user","ami-type":"AL2_x86_64","asg-min-size":30,"asg-max-size":35,"asg-desired-capacity":34,"image-id":"my-ami",  "ssm-document-create":true,   "instance-types":["type-2"],  "ssm-document-cfn-stack-name":"GetRef.Name-ssm", "ssm-document-name":"GetRef.Name-document",     "kubelet-extra-args":"aaa aa",  "volume-size":500}}`)
+	os.Setenv("AWS_K8S_TESTER_EKS_ADD_ON_NODE_GROUPS_ASGS", `{"GetRef.Name-ng-for-cni":{"name":"GetRef.Name-ng-for-cni","remote-access-user-name":"ec2-user","ami-type":"AL2_x86_64","asg-min-size":30,"asg-max-size":35,"asg-desired-capacity":34,"image-id":"my-ami",  "ssm-document-create":true,   "instance-types":["type-2"],  "ssm-document-cfn-stack-name":"GetRef.Name-ssm", "ssm-document-name":"GetRef.Name-document", "kubelet-extra-args":"aaa aa", "cluster_auto_scaler": {"enable" : false}, "volume-size":500}}`)
```
gyuho, line 1357: `cluster-autoscaler`

eksconfig/env_test.go
```diff
@@ -1426,7 +1429,7 @@ func TestEnvAddOnConformance(t *testing.T) {
 
 	os.Setenv("AWS_K8S_TESTER_EKS_ADD_ON_NODE_GROUPS_ENABLE", `true`)
 	defer os.Unsetenv("AWS_K8S_TESTER_EKS_ADD_ON_NODE_GROUPS_ENABLE")
-	os.Setenv("AWS_K8S_TESTER_EKS_ADD_ON_NODE_GROUPS_ASGS", `{"GetRef.Name-ng-for-cni":{"name":"GetRef.Name-ng-for-cni","remote-access-user-name":"ec2-user","ami-type":"AL2_x86_64","asg-min-size":30,"asg-max-size":35,"asg-desired-capacity":34,"image-id":"my-ami",  "ssm-document-create":true,   "instance-types":["type-2"],  "ssm-document-cfn-stack-name":"GetRef.Name-ssm", "ssm-document-name":"GetRef.Name-document",     "kubelet-extra-args":"aaa aa",  "volume-size":500}}`)
+	os.Setenv("AWS_K8S_TESTER_EKS_ADD_ON_NODE_GROUPS_ASGS", `{"GetRef.Name-ng-for-cni":{"name":"GetRef.Name-ng-for-cni","remote-access-user-name":"ec2-user","ami-type":"AL2_x86_64","asg-min-size":30,"asg-max-size":35,"asg-desired-capacity":34,"image-id":"my-ami",  "ssm-document-create":true,   "instance-types":["type-2"],  "ssm-document-cfn-stack-name":"GetRef.Name-ssm", "ssm-document-name":"GetRef.Name-document", "cluster_auto_scaler": {"enable" : false}, "kubelet-extra-args":"aaa aa",  "volume-size":500}}`)
```
gyuho, line 1432: `cluster-autoscaler`

eksconfig/add-on-node-groups.go
```diff
@@ -56,6 +56,14 @@ type AddOnNodeGroups struct {
 	ASGs map[string]ASG `json:"asgs,omitempty"`
 }
 
+type NGClusterAutoScaler struct {
+	Enable bool `json:"enable"`
+	// Created is true when the resource has been created.
+	// Used for delete operations.
+	Created         bool               `json:"created" read-only:"true"`
+	TimeFrameCreate timeutil.TimeFrame `json:"time-frame-create" read-only:"true"`
```
gyuho, line 64: Can we add another field `Image string`, and set default value.

eksconfig/README.md
hakuna-matatah: > We do not want to cluster-wide autoscaler enable. This should be scoped down to individual node groups. I think your code already does that. Can we remove this?

forgot t remove it.

eksconfig/add-on-node-groups.go
hakuna-matatah: We are not doing this for now. Giving that option would lead into other problems as supplying wrong image that's not in GCR etc.

eks/ng/autoscaler/cluster-autoscaler.go
```diff
@@ -0,0 +1,412 @@
...
+	}
+	return ts.createCA()
+}
+
+func (ts *tester) createCA() error {
+	ts.cfg.Logger.Info("creating CA using kubectl", zap.String("name", ts.cfg.EKSConfig.Name))
+	for asgName, cur := range ts.cfg.EKSConfig.AddOnNodeGroups.ASGs {
+		createStart := time.Now()
+		ts.cfg.Logger.Info("creating CA", zap.String("name", asgName))
+		if err := ts.installCA(); err != nil {
```
gyuho, line 268: `InstallCA` should take `cur.Name` as a parameter, otherwise we install the same plugin x number of times?

eks/ng/autoscaler/cluster-autoscaler.go
```diff
@@ -0,0 +1,412 @@
...
+			return err
+		}
+		ts.cfg.Logger.Info("created CA",
+			zap.String("asg-name", cur.Name),
+			zap.String("started", humanize.RelTime(createStart, time.Now(), "ago", "from now")),
+		)
+		ts.cfg.EKSConfig.AddOnNodeGroups.ASGs[asgName] = cur
+		_ = ts.cfg.EKSConfig.Sync()
```
gyuho, line 276: We can just ignore without `_=`?

eks/ng/autoscaler/cluster-autoscaler.go
```diff
@@ -0,0 +1,412 @@
...
+		createStart := time.Now()
+		ts.cfg.Logger.Info("creating CA", zap.String("name", asgName))
+		if err := ts.installCA(); err != nil {
+			return err
+		}
+		ts.cfg.Logger.Info("created CA",
+			zap.String("asg-name", cur.Name),
+			zap.String("started", humanize.RelTime(createStart, time.Now(), "ago", "from now")),
+		)
+		ts.cfg.EKSConfig.AddOnNodeGroups.ASGs[asgName] = cur
```
gyuho, line 275: Actually, remove line 275 and 276. Doesn't seem like we change anything in `cur`?

eks/ng/autoscaler/cluster-autoscaler.go
```diff
@@ -0,0 +1,412 @@
...
+		_ = ts.cfg.EKSConfig.Sync()
+	}
+	return ts.cfg.EKSConfig.Sync()
+}
+
+func (ts *tester) installCA() error {
+	var caData = caSpecData{}
+	caData.ImageURI = images[ts.cfg.EKSConfig.Parameters.Version]
```
gyuho, line 283:
```
caData.ImageURI, ok = images[...]
if !ok { return fmt.Errorf("no CA found for %q", ...Version) }
```

eks/ng/autoscaler/cluster-autoscaler.go
```diff
@@ -0,0 +1,412 @@
...
+	}
+	if err != nil {
+		return fmt.Errorf("'kubectl apply' failed %v (output %q)", err, string(output))
+	}
+	ts.cfg.Logger.Info("created cluster autoscaler")
+
+	return ts.waitDeploymentCA()
+}
+
+func (ts *tester) waitDeploymentCA() error {
```
gyuho, line 340: This should take `clusterAutoScalerDeploymentName` as a parameter, in case we run multiple cluster autoscalers for multiple node groups?

eks/ng/role.go
```diff
@@ -238,8 +257,14 @@ func (ts *tester) createRole() error {
 	if ts.cfg.EKSConfig.AddOnNodeGroups.RoleName == "" {
 		return errors.New("cannot create a cluster role with an empty AddOnNodeGroups.RoleName")
 	}
-
-	if err := ioutil.WriteFile(ts.cfg.EKSConfig.AddOnNodeGroups.RoleCFNStackYAMLFilePath, []byte(TemplateRole), 0400); err != nil {
+	tr := templateRole{}
+	tr.ASGPolicy = asgPolicyData
```
gyuho, line 261: Only set this when autoscaler is enabled?

eksconfig/add-on-node-groups.go
```diff
@@ -232,6 +243,10 @@ func (cfg *Config) validateAddOnNodeGroups() error {
 			return fmt.Errorf("%q both ImageID and ImageIDSSMParameter are empty", cur.Name)
 		}
 
+		if cur.ClusterAutoScaler.Enable && !cfg.AddOnNodeGroups.RoleCreate {
```
gyuho, line 246: Also error if `CA.Enable == true` and `AddOnMetricsServer.Enable == false`? Is metrics-server requirement for CA?

eks/ng/role.go
```diff
@@ -196,6 +198,7 @@ Resources:
             - route53:ChangeResourceRecordSets
             - route53:DeleteHealthCheck
             Resource: "*"
+{{ if ne .ASGPolicy "" }}{{.ASGPolicy}}{{ end }}
```
gyuho, line 201: Can we add a simple test when `ASGPolicy` is not empty?

https://github.com/aws/aws-k8s-tester/blob/3e23f03490ef60feceec33f37e78b1996f39f6b9/eks/cluster/cluster_test.go#L10

eks/ng/autoscaler/cluster-autoscaler.go
hakuna-matatah: > `InstallCA` should take `cur.Name` as a parameter, otherwise we install the same plugin x number of times?

I don't think it will install the same plugin. We are only installing once per cluster as long as one NG is enabled.

eks/ng/autoscaler/cluster-autoscaler.go
hakuna-matatah: My IDE was showing as a warning :D ...so just added it. I think that should be fine ?

eksconfig/add-on-node-groups.go
hakuna-matatah: > Also error if `CA.Enable == true` and `AddOnMetricsServer.Enable == false`? Is metrics-server requirement for CA?

metrics-server is not a requirement for CA. Metrics server is a requirement for HPA.

eks/ng/autoscaler/cluster-autoscaler.go
hakuna-matatah: cluster-autoscaler deployment will only be one per cluster.

eks/ng/role.go
hakuna-matatah: I think this should be ok ? Because by default these policies will be included to the role if the node-groups (or) clusters are created using "eksctl".

eks/ng/role.go
hakuna-matatah: > Can we add a simple test when `ASGPolicy` is not empty?
> 
> https://github.com/aws/aws-k8s-tester/blob/3e23f03490ef60feceec33f37e78b1996f39f6b9/eks/cluster/cluster_test.go#L10

Because we are adding ASGPolicy with value in any case, do we really care ?

eksconfig/add-on-node-groups.go
```diff
@@ -68,6 +76,8 @@ type ASG struct {
 	// TODO: handle conflicting flag '--cloud-provider aws'
 	// ref. https://github.com/kubernetes/kubernetes/issues/64659
 	KubeletExtraArgs string `json:"kubelet-extra-args"`
+	// clusterautoscaler per nodegroup
+	ClusterAutoScaler *NGClusterAutoScaler `json:"cluster-autoscaler,omitempty"`
```
gyuho, line 80: Let's move this inside `AddOnNodeGroups`

eks/ng/autoscaler/cluster-autoscaler.go
gyuho: Ok makes sense.

eks/ng/role.go
gyuho: ack

eksconfig/add-on-node-groups.go
gyuho: ack

eks/ng/autoscaler/cluster-autoscaler.go
gyuho: ok let's keep it then

eks/ng/autoscaler/cluster-autoscaler.go
gyuho: > for asgName, cur := range ts.cfg.EKSConfig.AddOnNodeGroups.ASGs {

still looping through this for-loop?  now that we are moving cluster autoscaler installer out, we can replace `createCA` with `installCA`?

gyuho: Talked offline. We still need per-ASG updates for tagging

eksconfig/add-on-node-groups.go
hakuna-matatah: talked offline. There is value in having it inside ASG. We can enable/disable CA per nodegroup.

PR #101, opened 2020-06-16
Remove duplication of colorized format strings

anguslees:
Use the "disable" feature of `colorstring` to avoid duplicating all
the format strings, just for the non-colored case.

Also enable the "reset" feature, so colored strings are automatically
reset at the end of each message.  This obsoletes the explicit
`[default]` statements, so remove those too.

### DIFF
--- a/ec2config/config.go
+++ b/ec2config/config.go
@@ -12,6 +12,7 @@ import (
 	"github.com/aws/aws-k8s-tester/pkg/timeutil"
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/mitchellh/colorstring"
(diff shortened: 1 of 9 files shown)


anguslees:
Some of this code is a bit weird, since I was minimising the change required to all the print statements.  At least it reduces it to one print statement per printed statement :sweat_smile:

ec2config/config.go
```diff
@@ -181,6 +182,11 @@ type Config struct {
 	ASGs map[string]ASG `json:"asgs"`
 }
 
+func (c Config) Colorize(input string) string {
+	colorize := colorstring.Colorize{Disable: !c.LogColor, Reset: true}
```
gyuho, line 186: nice!
didn't know about this :)

anguslees:
Oh, you probably want to revert that.  My mechanical translation reversed all the print lines - oops :stuck_out_tongue: 

Try this updated version.

PR #117, opened 2020-07-06
Add basic fluentd add-on

gyuho:


gyuho:
Renaming to container insights.

PR #118, opened 2020-07-06
*: allow custom busybox repository from ECR

gyuho:
In case dockerhub has outages, use ECR as fallback.

### DIFF
--- a/CHANGELOG/CHANGELOG-1.4.md
+++ b/CHANGELOG/CHANGELOG-1.4.md
@@ -3,7 +3,7 @@
 <hr>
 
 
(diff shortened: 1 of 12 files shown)


wongma7:
lgtm

Do you foresee the tests using any other images? It will be overkill for just one image, but if so, consider copying upstream's idea to organize all the necessary images by their registry https://github.com/kubernetes/kubernetes/blob/release-1.17/test/utils/image/manifest.go .

gyuho:
> Do you foresee the tests using any other images?

For now, just `busybox`.

> It will be overkill for just one image, but if so, consider copying upstream's idea to organize all the necessary images by their registry https://github.com/kubernetes/kubernetes/blob/release-1.17/test/utils/image/manifest.go .

Noted. Thanks!

Issue #123, opened 2020-07-14
Improve add-on testing experience with existing cluster


gyuho:
Cluster creation is expensive taking >20-min.

Create cluster once.
Run add-on tests.
Clean up add-on tests resources.
Run more add-on tests.
Clean up add-on tests resources.
(Optional) delete cluster.

ellistarn:
One possible user experience

A few key points
- This YAML serializes to a go struct, which represents the AKT configuration
- All keys are go pointers, and if omitted, it's effectively "disabled"
- Enables users to attach to existing test or cluster cluster
  - i.e., if an AKT cluster already exists, user can change addons config, and AKT will only apply the changes
  - A user developing AKT can make code changes to AKT and reapply the same config, which will simply apply again (if metadata.name is the same)
  - Each addon's installation and deletion must be idempotent
- There is a delete-cluster="false" option, so that addons can be uninstalled but the cluster remains

```      
AKT_CONFIG_FILE="" # User can either provide config file or yaml in line. If file not provided, AKT creates one to manage state inside "status"
AKT_CONFIG_YAML="
metadata: 
    name: cluster-name # Test name prefix, if not provided, generates name for test
spec: 
    eks: # Cluster creation fields
        region: us-west-2
        parititon: aws
    s3: 
        bucketName: # if not provided, creates a new bucket
    mng:
        groups: 
            - name: foo
              image-id-ssm-parameter
              instance-types: 
                - c5.xlarge
    clusterAutoscaler: {} # Enabled but uses defaults
    metricsServer: {} # Enabled but uses defaults
    irsa: 
        repositoryAccountId: 607362164682
        repositoryAccountName: aws/aws-k8s-tester
        repositoryImageTag: latest
status: {} # AKT maintains its state here
"
aws-k8s-tester eks create # Generates resources in AKT_CONFIG
aws-k8s-tester eks delete # Detects necessary info from yaml file 
```

PR #128, opened 2020-07-15
*: support "AWS_K8S_TESTER_EKS_SKIP_DELETE_CLUSTER_AND_NODES"

gyuho:
Partially address https://github.com/aws/aws-k8s-tester/issues/123.

Short-term solution without breaking changes.

### DIFF
--- a/CHANGELOG/CHANGELOG-1.4.md
+++ b/CHANGELOG/CHANGELOG-1.4.md
@@ -9,9 +9,23 @@
 
 See [code changes](https://github.com/aws/aws-k8s-tester/compare/v1.4.5...v1.4.6).
 
+### `eksconfig`
+
+- Add [`SkipDeleteClusterAndNodes` to skip EKS cluster and nodes deletion](https://github.com/aws/aws-k8s-tester/commit/565635179860b4832b64cbe3b39fdbe1c12b1ae1).
+  - Useful for testing add-ons with existing clusters.
+  - If true, delete operation keeps all resources created before cluster (e.g. IAM role, VPC, CMK, etc.).
+  - If true, delete operation keeps all node groups and managed node groups!
+  - Set via `AWS_K8S_TESTER_EKS_SKIP_DELETE_CLUSTER_AND_NODES=true`.
+  - Default `AWS_K8S_TESTER_EKS_SKIP_DELETE_CLUSTER_AND_NODES=false`.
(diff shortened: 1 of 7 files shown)


CHANGELOG/CHANGELOG-1.4.md
```diff
@@ -9,9 +9,20 @@
 
 See [code changes](https://github.com/aws/aws-k8s-tester/compare/v1.4.5...v1.4.6).
 
+### `eksconfig`
+
+- Add [`SkipDeleteClusterAndNodes` to skip EKS cluster and nodes deletion](https://github.com/aws/aws-k8s-tester/commit/565635179860b4832b64cbe3b39fdbe1c12b1ae1).
```
ellistarn, line 14: Consider calling this "Attach to cluster"

CHANGELOG/CHANGELOG-1.4.md
```diff
@@ -9,9 +9,20 @@
...
+  - Useful for testing add-ons with existing clusters.
+  - If true, delete operation keeps all resources created before cluster (e.g. IAM role, VPC, CMK, etc.).
+  - Set via `AWS_K8S_TESTER_EKS_SKIP_DELETE_CLUSTER_AND_NODES=true`.
+  - Default `AWS_K8S_TESTER_EKS_SKIP_DELETE_CLUSTER_AND_NODES=false`.
+  - `aws-k8s-tester eks create cluster --path` must be passed a valid configuration file with existing cluster information in order to use the existing cluster. Create cluster once, and delete all add-ons, and keep the cluster related fields in the YAML with all other fields removed.
```
ellistarn, line 19: with this path to yaml eventually be our custom resource yaml config?

CHANGELOG/CHANGELOG-1.4.md
ellistarn: Instead of skipping, can we idempotently check that it already exists and just succeed?

ellistarn: Thinking even higher level, how much effort would it be to make everything within AKT idempotent, so that we could modify the yaml config and it would simply execute the deltas?

gyuho: Do you mean skipping creation? Creation is already done here

https://github.com/aws/aws-k8s-tester/blob/4cad9b32c68f212265af8d825e586847104cfb50/eks/cluster/cluster.go#L288-L303

which means you can run `create cluster` command multiple times and it will automatically skip if already created.

CHANGELOG/CHANGELOG-1.4.md
gyuho: Yes.

PR #133, opened 2020-07-20
eks/*: allow duplicate create calls

gyuho:
c.f. https://github.com/aws/aws-k8s-tester/issues/123


gyuho:
On second thought, this kind of change is getting too manual... Seems better to go with generic wrapper for Create + Update calls:

https://github.com/aws/aws-k8s-tester/blob/master/pkg/k8s-client/k8s.go#L206-L225

https://github.com/aws/aws-k8s-tester/blob/8f65783d9628062bd7929ef54490ddba2ca9164f/pkg/k8s-client/k8s.go#L206-L225

PR #137, opened 2020-07-23
Implemented the ClusterAutoscaler Addon with kubemark compatibility

ellistarn:


eks/clusterautoscaler/clusterautoscaler.go
```diff
@@ -0,0 +1,113 @@
...
+
+// ClusterAutoscaler is an addon that installs Cluster Autoscaler
+type ClusterAutoscaler struct {
+	K8sClient k8sclient.EKS
+	Config    *eksconfig.Config
+}
+
+// IsEnabled returns true if enabled
+func (c *ClusterAutoscaler) IsEnabled() bool {
+	return c.Config.Spec.ClusterAutoscaler != nil
```
gyuho, line 26: I like the idea of `nil` check here without an extra `Enable` field. How are we going to integrate this to the current env parsing function? I'd like to keep env option in case one doesn't want to write an YAML.

eksconfig/spec.go
```diff
@@ -0,0 +1,7 @@
+package eksconfig
+
+// Spec contains specs for addons
+type Spec struct {
+	// ClusterAutoscaler defines config for the cluster autoscaler
+	ClusterAutoscaler *ClusterAutoscalerSpec `json:"clusterAutoscaler,omitempty"`
```
gyuho, line 6: `clusterAutoscaler` seems consistent with upstream Kubernetes, but makes it inconsistent with current `eksconfig` JSON field names used for env parsing.

pkg/util/gotemplate.go
```diff
@@ -0,0 +1,31 @@
+package gotemplate
```
gyuho, line 1: Should we rename the directory `pkg/util` to `pkg/gotemplate`?

eksconfig/clusterautoscaler.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+	spec.Image = spec.defaultImage()
+}
+
+func (spec *ClusterAutoscalerSpec) defaultImage() string {
+	if spec.Image != "" {
+		return spec.Image
+	}
+	if spec.CloudProvider == CloudProviderKubemark {
+		return "767520670908.dkr.ecr.us-west-2.amazonaws.com/cluster-autoscaler-kubemark:latest"
```
gyuho, line 53: For other add-ons, we specify account ID and repo name as separate fields:

https://github.com/aws/aws-k8s-tester/blob/354f0478ba82cf9a1c4fa890eda71956d6ff8c33/eksconfig/add-on-php-apache.go#L23-L33

in order to check if ECR image exists and prevent creating Pod with an invalid image. If it does not exist, we fail fast: before we create Pod and find out later.

What do you think?

I am also fine with this approach since it's simpler.

eks/clusterautoscaler/clusterautoscaler.go
gyuho: On second thought, `Enable` field is kind of needed, in case an add-on does not specify any other fields (e.g. NLB hello world just needs Enable == true with other fields set to defaults).

ellistarn: This can also be achieved using NLB = {}. Addon pointer existence can fill the role of "enabled"

Regarding integration with env variables, I was thinking that we could do the following order of operations.
1. Read provided YAML file into a go struct
2. Merge Addons from Env Vars into go struct
3. Call Default
4. Call Validate

WDYT?

eksconfig/spec.go
ellistarn: Yeah the goal here was to align with upstream kubernetes whereever possible. I also wasn't intending for these addons to be configurable via ENV vars and to only use env vars for the older addons. What do you think?

pkg/util/gotemplate.go
ellistarn: Is pkg effectively "util" for you? I agree that "util" is a dirty word since it doesn't have clear encapsulation, it's a catch-all. Perhaps we can discuss over chat.

eksconfig/clusterautoscaler.go
ellistarn: I was leaning away from strong validation. Typically I'd agree with fail fast, I'd only expect users to override image if they know what they're doing, and the additional complexity doesn't seem worth it.

eks/eks.go
```diff
@@ -1333,6 +1344,16 @@ func (ts *Tester) Up() (err error) {
 		fmt.Fprintf(ts.logWriter, "\nrunCommand output:\n\n%s\n", string(out))
 	}
 
+	// Generic mechanism for Creating addons. Add new addons to getInstallationOrder()
+	for _, order := range ts.getInstallationOrder() {
+		if err := ts.runAsync(order, func(a tester.Addon) error {
+			klog.Infof("Applying addon %s", reflect.TypeOf(a))
```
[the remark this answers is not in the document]
ellistarn: Oops, nice catch

eksconfig/spec.go
ellistarn: Also, https://github.com/iancoleman/strcase can easily convert camelCase or kabob-case into UPPER_SNAKE_CASE

gyuho: Good point. We can discuss this maybe in next iteration for full compatibilities.

PR #147, opened 2020-08-03
eksconfig: yaml.DisallowUnknownFields, parse EKS_CONFIG first

gyuho:


eksconfig/env_test.go
```diff
@@ -570,6 +580,10 @@ func TestEnv(t *testing.T) {
 		t.Fatal(err)
 	}
 
+	// TODO
+	fmt.Println(*cfg.Spec.ClusterAutoscaler)
```
ellistarn, line 584: Should you be doing an assert statement here instead of a print?

gyuho: Yeah now asserting with `testing.T`

PR #151, opened 2020-08-04
Refactor metrics-server to implement Addon interface

mirahmed753:
*Issue #, if available:*

*Description of changes:*
I refactored metrics-server to implement the Addon interface using overprovisioning and clusterautoscaler as references. Now, we can include `metrics-server` in the spec for `AWS_K8S_TESTER_EKS_CONFIG`.

### DIFF
--- a/eks/eks.go
+++ b/eks/eks.go
@@ -500,6 +500,10 @@ func (ts *Tester) createTesters() (err error) {
 			Config:    ts.cfg,
 			K8sClient: ts.k8sClient,
 		},
+		&metrics_server.MetricsServer{
+			Config:    ts.cfg,
+			K8sClient: ts.k8sClient,
+		},
 	}}
 
 	ts.testers = []eks_tester.Tester{
@@ -1361,6 +1365,7 @@ func (ts *Tester) Up() (err error) {
 		}); err != nil {
 			return fmt.Errorf("while applying addons, %w", err)
 		}
+		ts.cfg.Sync()
 	}
 
 	return ts.cfg.Sync()
--- a/eks/metrics-server/metrics-server.gotmpl
+++ b/eks/metrics-server/metrics-server.gotmpl
@@ -1,0 +1,164 @@
+# ref. https://docs.aws.amazon.com/eks/latest/userguide/dashboard-tutorial.html
+# ref. https://github.com/kubernetes-sigs/metrics-server/releases
+# ref. https://github.com/kubernetes-sigs/metrics-server/releases/download/v0.3.6/components.yaml
+---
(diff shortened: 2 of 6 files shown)


eks/eks.go
```diff
@@ -518,13 +522,6 @@ func (ts *Tester) createTesters() (err error) {
 			K8SClient: ts.k8sClient,
 			ECRAPI:    ecr.New(ts.awsSession, aws.NewConfig().WithRegion(ts.cfg.GetAddOnFluentdRepositoryBusyboxRegion())),
 		}),
-		metrics_server.New(metrics_server.Config{
```
ellistarn, line 521: Let's leave the old metrics serve in here. This is a backwards incompatible change, which will break all consumers (AWSWesleyCI). Typically, that way you want to do this is
1. Introduce the new changes
2. Migrate all consumers to the new way of doing things (this is called "deprecation")
3. Delete the old way of doing things.

eks/metrics-server/metrics-server.go
```diff
@@ -3,455 +3,56 @@
...
 
-// Config defines Dashboard configuration.
-type Config struct {
-	Logger    *zap.Logger
-	LogWriter io.Writer
-	Stopc     chan struct{}
-	EKSConfig *eksconfig.Config
-	K8SClient k8s_client.EKS
+// MetricsServer is an addon that installs Metrics Server
+type MetricsServer struct {
```
ellistarn, line 14: Easiest way to maintain backwards compat is by putting this new logic into metrics-server/addon.go and reverting this metrics-server.go file to what it was before.

eks/metrics-server/metrics-server.gotmpl
```diff
@@ -0,0 +1,164 @@
+# ref. https://docs.aws.amazon.com/eks/latest/userguide/dashboard-tutorial.html
```
ellistarn, line 1: Thanks for adding these!

eksconfig/metrics-server.go
```diff
@@ -0,0 +1,18 @@
+package eksconfig
+
+// MetricsServerSpec defines the spec for the Addon
```
ellistarn, line 3: I'm not sure if you noticed, but the way I wrote this comment makes it ultra copy-pastable :P

mirahmed753:
Made the changes based off of your comments. Let me know if I misunderstood anything :)

eks/eks.go
gyuho: Yeah let's see if we can keep two versions with shared code.

ellistarn:
LGTM!

PR #157, opened 2020-08-24
Added S3 Support for CL2 Tests

njtran:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/eks/cluster-loader/clusterloader2/addon.go
+++ b/eks/cluster-loader/clusterloader2/addon.go
@@ -1,7 +1,6 @@
 package clusterloader2
 
 import (
-	//"encoding/json"
 	"fmt"
 	"github.com/aws/aws-k8s-tester/eksconfig"
 	k8sclient "github.com/aws/aws-k8s-tester/pkg/k8s-client"
@@ -42,12 +41,11 @@ func (c *ClusterLoader) Apply() (err error) {
 	if err != nil {
 		return fmt.Errorf("while building templates, %v", err)
 	}
-	// Get Job info from goTemplate. Currently hardcoded.
+	// Get CL2 Job info from goTemplate
 	resources := strings.Split(template.String(), "\n---\n")
 	jobYaml := resources[len(resources)-1]
 	if err := c.K8sClient.Delete(jobYaml); err != nil {
 		//Warn that there was an error, but don't stop process.
-
 		zap.S().Warn("Deleting jobYaml errored out, but is fine: %s", zapcore.Field{String: err.Error()})
 	}
 	if err := c.K8sClient.Apply(template.String()); err != nil {
@@ -99,5 +97,6 @@ func (c *ClusterLoader) buildConfigMapData() (map[string]string, error) {
 }
 
 func (c *ClusterLoader) buildArgs() string {
-	return fmt.Sprintf("[%s]", strings.Join(c.Config.Spec.ClusterLoader.TestParams, ","))
+	// add report-dir argument for CL2 tests. If one wants to change, check the gotemplate, and look at the job's initContainer's volumes.
+	return fmt.Sprintf("[%s]", strings.Join(append(c.Config.Spec.ClusterLoader.TestParams, "--report-dir=/var/reports/cluster-loader"), ","))
 }
--- a/eks/cluster-loader/clusterloader2/cluster-loader.gotmpl
+++ b/eks/cluster-loader/clusterloader2/cluster-loader.gotmpl
@@ -46,17 +46,33 @@ metadata:
 spec:
   template:
     spec:
+      initContainers:
+        - name: clusterloader2
+          image: {{ .Image }}
+          command: ["/clusterloader"]
+          args: {{ .TestArgs }}
+          volumeMounts:
+            - name: config
+              mountPath: /etc/config
+            - name: reports
+              mountPath: /var/reports
+              readOnly: false
       containers:
-      - name: clusterloader2
-        image: {{ .Image }}
-        command: ["/clusterloader"]
-        args: {{ .TestArgs }}
-        volumeMounts:
-        - name: config
-          mountPath: /etc/config
+        - name: uploadresults
+          image: amazon/aws-cli:latest
+          command: 
+            - /bin/sh
+            - -c
+          args:
+            - aws s3 cp /var/reports/cluster-loader {{ .S3Uri }} --recursive        
+          volumeMounts: 
+            - name: reports
+              mountPath: /var/reports
       restartPolicy: Never
       serviceAccountName: clusterloader2
       volumes:
         - name: config
           configMap:
             name: clusterloader2-config
+        - name: reports
+          emptyDir: {}
--- a/eksconfig/cluster-loader.go
+++ b/eksconfig/cluster-loader.go
@@ -5,6 +5,7 @@ type ClusterLoaderSpec struct {
 	Image          string   `json:"image,omitempty"`
 	TestConfigUris []string `json:"testConfigUris,omitempty"`
 	TestParams     []string `json:"testParams,omitempty"`
+	S3Uri          string   `json:"s3Uri,omitempty"`
 }
 
 // MetricsServerStatus defines the status for the Addon


eks/cluster-loader/clusterloader2/addon.go
```diff
@@ -99,5 +99,17 @@ func (c *ClusterLoader) buildConfigMapData() (map[string]string, error) {
 }
 
 func (c *ClusterLoader) buildArgs() string {
-	return fmt.Sprintf("[%s]", strings.Join(c.Config.Spec.ClusterLoader.TestParams, ","))
+	return fmt.Sprintf("[%s]", strings.Join(append(c.Config.Spec.ClusterLoader.TestParams, "--report-dir=/usr/share/cluster-loader"), ","))
```
ellistarn, line 102: This is going to be marginally difficult to extend for future engineers. I might do something like:
```
arguments := c.Config.Spec.ClusterLoader.TestParams
for _, argument := []string{
  "--report-dir=/usr/share/cluster-loader",
} {
  arguments = append(arguments, argument)
}
return fmt.Sprintf("[%s]", strings.Join(arguments, ","))
```

ellistarn, line 102: nit: What led you to use /usr/share? I might use something like /var/reports.
https://tldp.org/LDP/Linux-Filesystem-Hierarchy/html/var.html

eks/cluster-loader/clusterloader2/addon.go
```diff
@@ -99,5 +99,17 @@ func (c *ClusterLoader) buildConfigMapData() (map[string]string, error) {
...
+}
+
+func (c *ClusterLoader) buildS3Command() string {
```
JacobGabrielson, line 105: I think it might be a little easier to read the script if it were a go template and you filled in the blanks? Would that be possible? Not a huge deal if it's too big of a change.

eks/cluster-loader/clusterloader2/cluster-loader.gotmpl
```diff
@@ -1,14 +1,14 @@
 apiVersion: v1
 kind: Namespace
 metadata:
-  name: clusterloader2
+  name: cl2
```
ellistarn, line 4: I prefer the more verbose clusterloader2 namespace. If we establish a pattern where addon name and namespace of the resources have parity, that will make things easier to work with.

eks/cluster-loader/clusterloader2/cluster-loader.gotmpl
```diff
@@ -42,21 +42,34 @@ apiVersion: batch/v1
...
+          image: {{ .Image }}
+          command: ["/clusterloader"]
+          args: {{ .TestArgs }}
+          volumeMounts:
+            - name: config
+              mountPath: /etc/config
+            - name: reports
+              mountPath: /usr/share
+              readOnly: false
+        - name: results
```
ellistarn, line 60: consider calling this "upload results"

eks/cluster-loader/clusterloader2/cluster-loader.gotmpl
```diff
@@ -42,21 +42,34 @@ apiVersion: batch/v1
...
+          image: amazon/aws-cli:latest
+          command: ["/bin/bash"]
+          args: {{ .S3Command }}
```
ellistarn, line 63: I might imbed the entire script you've created into the go template. Right now, it's difficult to understand what's going on in the golang code with a cursory read. 

The only variable you have is .S3URI, so you could just substitute that in the gotemplate.

eks/cluster-loader/clusterloader2/cluster-loader.gotmpl
```diff
@@ -42,21 +42,34 @@ apiVersion: batch/v1
...
+          volumeMounts: 
+            - name: reports
+              mountPath: /var/reports
+              readOnly: false
```
ellistarn, line 67: Doesn't this container only need readonly access?

eks/cluster-loader/clusterloader2/cluster-loader.gotmpl
```diff
@@ -42,21 +42,34 @@ apiVersion: batch/v1
...
-          mountPath: /etc/config
+        - name: clusterloader2
+          image: {{ .Image }}
+          command: ["/clusterloader"]
+          args: {{ .TestArgs }}
+          volumeMounts:
+            - name: config
+              mountPath: /etc/config
+            - name: reports
+              mountPath: /usr/share
```
ellistarn, line 58: why not /usr/reports?

eks/cluster-loader/clusterloader2/addon.go
```diff
@@ -99,5 +99,17 @@ func (c *ClusterLoader) buildConfigMapData() (map[string]string, error) {
...
+	// Separating large bash command for readability
+	bashVars := "sizeoffiles=$(ls /var/reports/cluster-loader -l | /bin/bash -c 'wc -l') ; initial=$sizeoffiles"
+
+	// have condition be add 2 because only keeping track of 2 metrics in my tests, so directory will be at least 2 files larger after test
```
ellistarn, line 109: If I understand it, this variable will change if we keep track of multiple metrics (i.e. multiple files). Let's chat offline to brainstorm ways that we can determine if CL2 is done or not.

eks/cluster-loader/clusterloader2/cluster-loader.gotmpl
```diff
@@ -42,21 +42,34 @@ apiVersion: batch/v1
...
+              readOnly: false
+        - name: results
+          image: amazon/aws-cli:latest
+          command: ["/bin/bash"]
```
JacobGabrielson, line 62: If possible, I'd use "/bin/sh" here and remove any Bash-isms ... Bash is a pretty complex beast and I think for these kind of automation situation's it's better to stay vanilla

eks/cluster-loader/clusterloader2/addon.go
njtran: I was in the habit of making it fully configurable in go, but it does make more sense to put most of it in the gotemplate. Will do.

eks/cluster-loader/clusterloader2/addon.go
njtran: Ah! Good idea.

njtran: Maybe I had some weird testing configurations, but I thought only /usr/share had worked. Now that I know a little more, I can see if /var/reports would work.

eks/cluster-loader/clusterloader2/cluster-loader.gotmpl
njtran: I originally made it cl2 for easier debugging purposes, but I like the idea of standardizing the namespaces.

eks/cluster-loader/clusterloader2/cluster-loader.gotmpl
njtran: Sounds good to me

eks/cluster-loader/clusterloader2/cluster-loader.gotmpl
njtran: As per Jacob's comment at the top, will do.

eks/cluster-loader/clusterloader2/cluster-loader.gotmpl
njtran: True! Good catch.

eks/cluster-loader/clusterloader2/cluster-loader.gotmpl
njtran: Fair. Will change.

eks/cluster-loader/clusterloader2/cluster-loader.gotmpl
njtran: Good idea. Will do.

eks/cluster-loader/clusterloader2/addon.go
```diff
@@ -1,7 +1,6 @@
 package clusterloader2
 
 import (
-	//"encoding/json"
```
ellistarn, line 4: dd

eks/cluster-loader/clusterloader2/addon.go
njtran: Synced offline, will use initContainers instead.

eks/cluster-loader/clusterloader2/addon.go
njtran: I'm going to leave as is. The appending argument isn't something that's going to need changing.

eks/cluster-loader/clusterloader2/cluster-loader.gotmpl
```diff
@@ -46,17 +46,34 @@ metadata:
...
-        args: {{ .TestArgs }}
-        volumeMounts:
-        - name: config
-          mountPath: /etc/config
+        - name: uploadresults
+          image: amazon/aws-cli:latest
+          command: 
+            - /bin/sh
+            - -c
+            - >
```
ellistarn, line 66: No need to do ">". That's for multiline strings. Your command is a one liner.
I'd recommend:
```
command:
  - /bin/sh
  - -c
arguments:
  -  aws s3 cp /var/reports/cluster-loader {{ .S3Uri }} --recursive
```
That will make everything but arguments very copy pastable.

eks/cluster-loader/clusterloader2/cluster-loader.gotmpl
```diff
@@ -46,17 +46,34 @@ metadata:
...
+              aws s3 cp /var/reports/cluster-loader {{ .S3Uri }} --recursive        
+          volumeMounts: 
+            - name: reports
+              mountPath: /var/reports
+              readOnly: true
```
ellistarn, line 71: What's the default on this field? I'd omit readOnly if the value you set is the same as the default.

njtran: Fair enough.

eks/cluster-loader/clusterloader2/cluster-loader.gotmpl
njtran: Got it.

PR #158, opened 2020-08-25
CL2 TestOverrides capabilities added

njtran:
*Issue #, if available:*

*Description of changes:*
In clusterloader2, there are some basic flags that drive the shared functionality between different test configurations, such as --nodes or --report-dir, but to further configure your test, you'd need to change your variables in the yaml, such as having a pods_per_node, num_of_namespaces, or enable_chaosmonkey variables. This allows total automation of tests, and by adding override fields, you can configure the test just as well as if it was in the CLI.

### DIFF
--- a/eks/cluster-loader/clusterloader2/addon.go
+++ b/eks/cluster-loader/clusterloader2/addon.go
@@ -13,6 +13,8 @@ import (
 	"strings"
 )
 
+const IndentedNewline = "\n    "
+
 type ClusterLoader struct {
 	Config    *eksconfig.Config
 	K8sClient k8sclient.EKS
@@ -79,6 +81,7 @@ func (c *ClusterLoader) Delete() (err error) {
 }
 
 func (c *ClusterLoader) buildConfigMapData() (map[string]string, error) {
+	// Any file in this map is at /etc/config in the mounted volume of CL2
 	dataMap := make(map[string]string)
 	for _, uri := range c.Config.Spec.ClusterLoader.TestConfigUris {
 		resp, err := http.Get(uri)
@@ -90,13 +93,21 @@ func (c *ClusterLoader) buildConfigMapData() (map[string]string, error) {
 			return nil, fmt.Errorf("while reading downloaded uri %s content: %v", uri, err)
 		}
 		// Indents the content by four spaces for valid YAML formatting
-		bodyString := strings.ReplaceAll(string(body), "\n", "\n    ")
+		bodyString := strings.ReplaceAll(string(body), "\n", IndentedNewline)
 		dataMap[path.Base(uri)] = bodyString
+	}
+	overrides := c.Config.Spec.ClusterLoader.TestOverrides
+	if len(overrides) > 0 {
+		dataMap["overrides.yaml"] = strings.Join(overrides, IndentedNewline)
 	}
 	return dataMap, nil
 }
 
 func (c *ClusterLoader) buildArgs() string {
-	// add report-dir argument for CL2 tests. If one wants to change, check the gotemplate, and look at the job's initContainer's volumes.
-	return fmt.Sprintf("[%s]", strings.Join(append(c.Config.Spec.ClusterLoader.TestParams, "--report-dir=/var/reports/cluster-loader"), ","))
+	// add report-dir and testOverrides for CL2 tests. If one wants to change, check the gotemplate, and look at the job's initContainer's volumes
+	arguments := append(c.Config.Spec.ClusterLoader.TestParams, "--report-dir=/var/reports/cluster-loader")
+	if len(c.Config.Spec.ClusterLoader.TestOverrides) > 0 {
+		arguments = append(arguments, "--testoverrides=/etc/config/overrides.yaml")
+	}
+	return fmt.Sprintf("[%s]", strings.Join(arguments, ","))
 }
--- a/eksconfig/cluster-loader.go
+++ b/eksconfig/cluster-loader.go
@@ -6,6 +6,7 @@ type ClusterLoaderSpec struct {
 	TestConfigUris []string `json:"testConfigUris,omitempty"`
 	TestParams     []string `json:"testParams,omitempty"`
 	S3Uri          string   `json:"s3Uri,omitempty"`
+	TestOverrides  []string `json:"testOverrides,omitempty"`
 }
 
 // MetricsServerStatus defines the status for the Addon


JacobGabrielson:
lgtm

ellistarn:
What is the use case for this?

ellistarn:
My understanding is that we drive tests via a couple mechanisms:
1. testparams: this lets us parameterize any test
2. testuri: this lets us point at tests that may look completely different.

What case are you trying to cover that isn't covered by these tests? What overrides are you trying to apply?

njtran:
> My understanding is that we drive tests via a couple mechanisms:
> 
> 1. testparams: this lets us parameterize any test
> 2. testuri: this lets us point at tests that may look completely different.
> 
> What case are you trying to cover that isn't covered by these tests? What overrides are you trying to apply?

Within a testUri there might be something like the following line https://github.com/aws/aws-k8s-tester/blob/master/eks/cluster-loader/configs/cluster-autoscaler/config.yaml#L3 which dictates a variable to a default value, but only if there is no go variable by the name of NAMESPACES. This allows the test to read in variables and further change the test past what is implemented in base options for flags.

Here is what perf-tests/clusterloader describes overrides as:
Overrides allow to inject new variables values to the template.
Many tests define input parameters. Input parameter is a variable that potentially will be provided by the test framework. Cause input parameters are optional, each reference has to be opaqued with DefaultParam function that will handle case if given variable doesn't exist.

ellistarn:
Isn't this currently possible by passing these values via ClusterLoader.TestParams?

eks/cluster-loader/clusterloader2/addon.go
```diff
@@ -93,10 +93,19 @@ func (c *ClusterLoader) buildConfigMapData() (map[string]string, error) {
 		bodyString := strings.ReplaceAll(string(body), "\n", "\n    ")
 		dataMap[path.Base(uri)] = bodyString
 	}
+	if len(c.Config.Spec.ClusterLoader.TestOverrides) > 0 {
+		overrides := strings.Join(c.Config.Spec.ClusterLoader.TestOverrides, "\n    ")
+		overrides = strings.ReplaceAll(overrides, "=", ": ")
+		dataMap["overrides.yaml"] = overrides
```
ellistarn, line 99: Is there any harm in letting dataMap[overrides.yaml] = "". That could simplify your if check.

eks/cluster-loader/clusterloader2/addon.go
```diff
@@ -93,10 +93,19 @@ func (c *ClusterLoader) buildConfigMapData() (map[string]string, error) {
 		bodyString := strings.ReplaceAll(string(body), "\n", "\n    ")
 		dataMap[path.Base(uri)] = bodyString
 	}
+	if len(c.Config.Spec.ClusterLoader.TestOverrides) > 0 {
+		overrides := strings.Join(c.Config.Spec.ClusterLoader.TestOverrides, "\n    ")
```
ellistarn, line 97: You're using this indentation in multiple places. Can you extract out a constant IndentedNewline = "\n    "

eks/cluster-loader/clusterloader2/addon.go
```diff
@@ -93,10 +93,19 @@ func (c *ClusterLoader) buildConfigMapData() (map[string]string, error) {
...
+		overrides = strings.ReplaceAll(overrides, "=", ": ")
```
ellistarn, line 98: Why not just have the user specify the correct value?

eks/cluster-loader/clusterloader2/addon.go
```diff
@@ -93,10 +93,19 @@ func (c *ClusterLoader) buildConfigMapData() (map[string]string, error) {
 		bodyString := strings.ReplaceAll(string(body), "\n", "\n    ")
 		dataMap[path.Base(uri)] = bodyString
 	}
+	if len(c.Config.Spec.ClusterLoader.TestOverrides) > 0 {
```
ellistarn, line 96: You could probably oneline this whole thing like 
```
dataMap["overrides.yaml"] = strings.Join(c.Config.Spec.ClusterLoader.TestOverrides, IndentedNewline)
```

eks/cluster-loader/clusterloader2/addon.go
```diff
@@ -93,10 +93,19 @@ func (c *ClusterLoader) buildConfigMapData() (map[string]string, error) {
...
 	return dataMap, nil
 }
 
 func (c *ClusterLoader) buildArgs() string {
-	// add report-dir argument for CL2 tests. If one wants to change, check the gotemplate, and look at the job's initContainer's volumes.
-	return fmt.Sprintf("[%s]", strings.Join(append(c.Config.Spec.ClusterLoader.TestParams, "--report-dir=/var/reports/cluster-loader"), ","))
+	// add report-dir and testOverrides for CL2 tests. If one wants to change, check the gotemplate, and look at the job's initContainer's volumes
+	arguments := append(c.Config.Spec.ClusterLoader.TestParams, "--report-dir=/var/reports/cluster-loader")
+	if len(c.Config.Spec.ClusterLoader.TestOverrides) > 0 {
+		arguments = append(arguments, "--testoverrides=/etc/config/overrides.yaml")
```
ellistarn, line 108: If you always have overrides (and just let it be empty if empty), you don't need an if check here.

eksconfig/cluster-loader.go
```diff
@@ -6,6 +6,7 @@ type ClusterLoaderSpec struct {
 	TestConfigUris []string `json:"testConfigUris,omitempty"`
 	TestParams     []string `json:"testParams,omitempty"`
 	S3Uri          string   `json:"s3Uri,omitempty"`
+	TestOverrides  []string `json:"testOverrides,omitempty"`
```
ellistarn, line 9: I'd strongly recommend either a string (cl2's convention) or a map (best go data type to represent the data). What you're doing is inventing a new data type to represent a file and then hardcoding some newlines into it.

eks/cluster-loader/clusterloader2/addon.go
njtran: Good idea

eks/cluster-loader/clusterloader2/addon.go
njtran: Will change.

eks/cluster-loader/clusterloader2/addon.go
njtran: I don't believe so, if there isn't harm, I'll change it.

eks/cluster-loader/clusterloader2/addon.go
njtran: gotcha. quick and easy.

eks/cluster-loader/clusterloader2/addon.go
njtran: true, will see if it is fine.

eksconfig/cluster-loader.go
njtran: I think having an array as input would make more sense to the users as opposed to having them format their overrides into one string, or even as a map. Having it as an array of strings is more akin to how one would create their own overrides file for CL2, which I would argue makes the most sense.

ellistarn: Ok, SGTM.

eks/cluster-loader/clusterloader2/addon.go
njtran: Returns an error, keeping the check.

eks/cluster-loader/clusterloader2/addon.go
njtran: Can't have an empty testoverrides.

eks/cluster-loader/clusterloader2/addon.go
```diff
@@ -78,7 +78,10 @@ func (c *ClusterLoader) Delete() (err error) {
 	return nil
 }
 
+// Any file in this map is at /etc/config in the mounted volume of CL2
 func (c *ClusterLoader) buildConfigMapData() (map[string]string, error) {
+	const indentedNewline = "\n    "
```
ellistarn, line 83: Make this a file level constant.

PR #161, opened 2020-08-31
Added warning message for configuration file, and fixed a minor bug

njtran:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/eks/cluster-loader/configs/cluster-autoscaler/config.yaml
+++ b/eks/cluster-loader/configs/cluster-autoscaler/config.yaml
@@ -1,4 +1,6 @@
-# Adjust the following two for your cluster -
+##### README #####
+# If (nodes*pods_per_node*client_qps % (pods_per_second*namespaces) != 0), test will fail.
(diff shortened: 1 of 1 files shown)


eks/cluster-loader/configs/cluster-autoscaler/config.yaml
```diff
@@ -1,3 +1,5 @@
+# Adjust the following two for your cluster -
+### CAREFUL -- if (nodes*pods_per_node*client_qps % (pods_per_second*namespaces) != 0), test will fail. Ensure your parameters meet this condition.
```
ellistarn, line 2: What do you mean by adjust the following two for your cluster? 

I'd also just use a single comment, unless you want to do something like:

##### README #####
# If (nodes*pods_per_node*client_qps % (pods_per_second*namespaces) != 0), test will fail. 
# This is a clusterloader2 restriction, so ensure your parameters meet this condition.

njtran: Got it. I like this idea.

PR #174, opened 2020-10-05
Changed CL2 test add-on to not retry on failure, and fixed CA add-on max nodes error

njtran:
*Issue #, if available:*

*Description of changes:*
- Added backoffLimit so that CL2 addon test doesn't retry
- Changed default deployment.yaml to be real nodes (not hollow-nodes)
- Added hollowdeployment.yaml for option for hollow-nodes
- Fixed a CA add-on error where maxNodes was mis-coded as minNodes.

### DIFF
--- a/eks/cluster-loader/clusterloader2/cluster-loader.gotmpl
+++ b/eks/cluster-loader/clusterloader2/cluster-loader.gotmpl
@@ -44,6 +44,7 @@ metadata:
   name: clusterloader2
   namespace: clusterloader2
 spec:
+  backoffLimit: 0
   template:
     spec:
       initContainers:
--- a/eks/cluster-loader/configs/cluster-autoscaler/hollowdeployment.yaml
+++ b/eks/cluster-loader/configs/cluster-autoscaler/hollowdeployment.yaml
@@ -23,6 +23,15 @@ spec:
             requests:
               cpu: {{.CpuRequest}}
               memory: {{.MemoryRequest}}
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+              - matchExpressions:
+                  - key: NodeType
+                    operator: In
+                    values:
+                      - hollow-nodes
       # Add not-ready/unreachable tolerations for 15 minutes so that node
       # failure doesn't trigger pod deletion.
       tolerations:
@@ -34,3 +43,7 @@ spec:
           operator: "Exists"
           effect: "NoExecute"
           tolerationSeconds: 900
+        - key: "provider"
+          value: "kubemark"
+          operator: "Equal"
+          effect: "NoSchedule"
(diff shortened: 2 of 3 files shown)


eks/cluster-loader/configs/cluster-autoscaler/hollowdeployment.yaml
```diff
@@ -0,0 +1,49 @@
+apiVersion: apps/v1
```
ellistarn, line 1: This file doesn't look like its used anywhere.

eks/cluster-loader/configs/cluster-autoscaler/deployment.yaml
```diff
@@ -23,15 +23,6 @@ spec:
             requests:
               cpu: {{.CpuRequest}}
               memory: {{.MemoryRequest}}
-      affinity:
```
ellistarn, line 26: What's the value in removing this affinity and the toleraiton?

eks/cluster-loader/clusterloader2/cluster-loader.gotmpl
```diff
@@ -44,6 +44,7 @@ metadata:
   name: clusterloader2
   namespace: clusterloader2
 spec:
+  backoffLimit: 0
```
ellistarn, line 47: What does this do?

njtran: If the test doesn't pass for whatever reason, without this, it will attempt to restart the init container. Restarting the init container will do another CL2 test, which we don't want.

eks/cluster-loader/configs/cluster-autoscaler/hollowdeployment.yaml
njtran: The point was to have the necessary changes to make it kubemark possible if one would like to switch back.

eks/cluster-loader/clusterloader2/cluster-loader.gotmpl
ellistarn: Nice!

eks/cluster-loader/configs/cluster-autoscaler/hollowdeployment.yaml
ellistarn: Ah -- this is what git history is for. I think we should leave the old CA as is. It's compatible with both hollow and real nodes.

ellistarn: Oh -- wait, the affinity is required. You could change
requiredDuringSchedulingIgnoredDuringExecution
preferredDuringSchedulingIgnoredDuringExecution

which would keep compatibility with both approaches.

The key is that you dont want to add a new feature while breaking an old one.

eks/cluster-loader/configs/cluster-autoscaler/deployment.yaml
ellistarn: Discussed offline. This is fine.

ellistarn (approved):
Ignore me I was confused.

eks/cluster-loader/configs/cluster-autoscaler/deployment.yaml
njtran: toleration not needed if it's not kubemark, and same goes with the affinity

PR #175, opened 2020-10-07
Changed VPC defaults to be multiple blocks to account for larger testing clusters

njtran:
*Issue #, if available:*

*Description of changes:*
- Expanded VPC default CIDR range in order to support more pods for larger scale tests
  - Previous VPC defaults had one /19 CIDR Block, allowing for 8k pods. Added multiple blocks of max VPC Block size (/16).
  - Changed VPCs from 192 space to 10 space.
- Subnets are by default same CIDR range as VPC Blocks, but can be changed with environment variables
  - Public Subnets are /16 blocks by default
  - Private Subnets are /17 blocks by default

*Testing*
- Tested by bringing up a cluster to a larger size than before to see if VPC was able to handle more IPs than before the change. VPC was brought up successfully, and more IPs were successfully utilized.

### DIFF
--- a/eks/cluster-loader/clusterloader2/addon.go
+++ b/eks/cluster-loader/clusterloader2/addon.go
@@ -1,23 +1,17 @@
 package clusterloader2
 
 import (
-	"context"
 	"fmt"
 	"io/ioutil"
 	"net/http"
 	"path"
 	"strings"
-	"time"
 
 	"github.com/aws/aws-k8s-tester/eksconfig"
 	k8sclient "github.com/aws/aws-k8s-tester/pkg/k8s-client"
 	gotemplate "github.com/aws/aws-k8s-tester/pkg/util"
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
-	v1 "k8s.io/api/batch/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/client-go/util/retry"
 )
 
 // IndentedNewline covers formatting issues with gotemplates
@@ -44,10 +38,12 @@ func (c *ClusterLoader) Apply() (err error) {
 		*eksconfig.ClusterLoaderSpec
 		ConfigMapData map[string]string
 		TestArgs      string
+		InstanceTypes []string
 	}{
 		ClusterLoaderSpec: c.Config.Spec.ClusterLoader,
 		ConfigMapData:     configMapData,
 		TestArgs:          c.buildArgs(),
+		InstanceTypes:     c.Config.Spec.ClusterLoader.InstanceTypes,
 	})
 	if err != nil {
 		return fmt.Errorf("while building templates, %v", err)
@@ -67,20 +63,6 @@ func (c *ClusterLoader) Apply() (err error) {
 			Installed: true,
 			Ready:     true,
 		},
-	}
-
-	// Wait for job to complete -- 2 hours because larger tests take a very long time.
-	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Hour)
-	defer cancel()
-	job := &v1.Job{}
-	for job.Status.Succeeded < 1 {
-		if err := retry.OnError(retry.DefaultRetry, errors.IsTimeout, func() (err error) {
-			job, err = c.K8sClient.KubernetesClientSet().BatchV1().Jobs("clusterloader2").Get(ctx, "clusterloader2", metav1.GetOptions{})
-			return err
-		}); err != nil {
-			return fmt.Errorf("failed to get cl2 job (%v)", err)
-		}
-		time.Sleep(10 * time.Second)
 	}
 	return nil
 }
@@ -135,7 +117,6 @@ func (c *ClusterLoader) buildArgs() string {
 		"--provider=eks",
 		"--testconfig=/etc/config/config.yaml",
 		"--run-from-cluster",
-		"--report-dir=/var/reports/cluster-loader",
 		"--alsologtostderr",
 	} {
 		arguments = append(arguments, argument)
--- a/eks/cluster/vpc.go
+++ b/eks/cluster/vpc.go
@@ -83,40 +83,58 @@ Parameters:
     Type: String
     Default: aws-k8s-tester-eks-vpc
 
-  VPCCIDR:
-    Description: IP range (CIDR notation) for VPC, must be a valid (RFC 1918) CIDR range (from 192.168.0.0 to 192.168.255.255)
+  VPCBlock1: 
     Type: String
-    Default: 192.168.0.0/16
+    Default: 10.0.0.0/16
+    Description: The primary CIDR range for the VPC. This should be a valid private (RFC 1918) CIDR range.
+    AllowedPattern: '((\d{1,3})\.){3}\d{1,3}/\d{1,2}'
+
+  VPCBlock2:
+    Type: String
+    Default: 10.1.0.0/16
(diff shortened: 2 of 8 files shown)


eks/cluster-loader/clusterloader2/addon.go
```diff
@@ -68,20 +63,6 @@ func (c *ClusterLoader) Apply() (err error) {
 			Ready:     true,
 		},
 	}
-
-	// Wait for job to complete -- 2 hours because larger tests take a very long time.
-	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Hour)
-	defer cancel()
-	job := &v1.Job{}
-	for job.Status.Succeeded < 1 {
```
gyuho, line 76: Can you explain why we are getting rid of this? It just waits for at least one job is successful...

eks/cluster/vpc.go
```diff
@@ -567,9 +601,19 @@ Outputs:
     Value: !Ref VPC
     Description: VPC ID
 
-  VPCCIDR:
-    Value: !Ref VPCCIDR
-    Description: VPCCIDR
+  VPCBlock1:
+    Value: !Ref VPCBlock1
+    Description: Primary VPC Block
+
```
gyuho, line 607: Remove this?

eks/cluster/vpc.go
```diff
@@ -567,9 +601,19 @@ Outputs:
...
+  VPCBlock2:
+    Value: !Ref VPCBlock2
+    Description: Primary VPC Block
```
gyuho, line 610: s/Primary/Secondary/?

gyuho (changes requested):
`PublicSubnetCIDR*` and `VPCBlock*` are the same by default. Can we just use VPC block for subnet CIDR ranges?

eks/cluster-loader/clusterloader2/addon.go
```diff
@@ -1,23 +1,16 @@
...
-
 	"github.com/aws/aws-k8s-tester/eksconfig"
 	k8sclient "github.com/aws/aws-k8s-tester/pkg/k8s-client"
 	gotemplate "github.com/aws/aws-k8s-tester/pkg/util"
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
-	v1 "k8s.io/api/batch/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/client-go/util/retry"
```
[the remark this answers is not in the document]
njtran: Will do.

eks/cluster-loader/clusterloader2/addon.go
njtran: We were going to use it for when we put certain tests into Prow, but found that this didn't properly handle the timeouts. Rather than trying to figure out the timeouts, I decided to get rid of it as we didn't need it, and may not need it later.

eks/cluster/vpc.go
njtran: Are you referring to the newline or the VPCBlock1?

eks/cluster/vpc.go
njtran: Good catch.

eks/cluster/vpc.go
gyuho: Empty new line

eks/cluster-loader/clusterloader2/addon.go
gyuho: > didn't properly handle the timeouts

Then, we can decrease the timeouts and just log in warning level? Do you run clusterloader2 for more than 2 hours?

gyuho:
Let's document in CHANGELOG, we are now adding default VPC blocks with the same range as subnet CIDRs. And highlight default blocks are expanded to support large scale workloads.

eks/cluster-loader/clusterloader2/addon.go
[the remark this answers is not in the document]
ellistarn: Your editor should automate ths. Talk to me if you need help setting this up.

eks/cluster-loader/clusterloader2/addon.go
```diff
@@ -44,10 +37,12 @@ func (c *ClusterLoader) Apply() (err error) {
 		*eksconfig.ClusterLoaderSpec
 		ConfigMapData map[string]string
 		TestArgs      string
+		InstanceTypes []string
 	}{
 		ClusterLoaderSpec: c.Config.Spec.ClusterLoader,
 		ConfigMapData:     configMapData,
 		TestArgs:          c.buildArgs(),
+		InstanceTypes:     c.Config.Spec.ClusterLoader.InstanceTypes,
```
ellistarn, line 45: IMO instance types shouldn't be used to select. Perhaps we can use a node selector?

eks/cluster/vpc.go
```diff
@@ -836,11 +898,26 @@ func (ts *tester) createVPC() error {
...
-		case "VPCCIDR":
+		case "VPCBlock1":
 			// to avoid additional CFN describe stack call
-			// to look for VPCCIDR parameter value
+			// to look for VPCBlock1 parameter value
 			// ref. https://github.com/aws/aws-k8s-tester/pull/162
-			ts.cfg.EKSConfig.Parameters.VPCCIDR = v
+			ts.cfg.EKSConfig.Parameters.VPCBlock1 = v
+		case "VPCBlock2":
+			// to avoid additional CFN describe stack call
```
ellistarn, line 907: Don't copy this comment 4 times.

ellistarn (commented):
Thanks for the cleanup. A couple comments

eks/cluster-loader/clusterloader2/addon.go
njtran: Depending on the size of the test, yes it can run for up to two hours. But when scale tests get too large, API calls time out, and the testing for that takes a very long time, so rather than having a partially working feature, I just wanted to remove it.

eks/cluster/vpc.go
njtran: +1

eks/cluster-loader/clusterloader2/addon.go
gyuho: > Depending on the size of the test, yes it can run for up to two hours. But when scale tests get too large, API calls time out, and the testing for that takes a very long time, so rather than having a partially working feature, I just wanted to remove it.

I understand that. Still, this breaks other small-scale tests where 2-hour is enough. Our bug bash job relies on this checks to make sure 5 rounds of clusterloader runs complete successfully.

We can add another field to `eksconfig` to opt out of this.

eks/cluster-loader/clusterloader2/addon.go
```diff
@@ -135,7 +117,6 @@ func (c *ClusterLoader) buildArgs() string {
 		"--provider=eks",
 		"--testconfig=/etc/config/config.yaml",
 		"--run-from-cluster",
-		"--report-dir=/var/reports/cluster-loader",
```
gyuho, line 138: Why do we remove this?

njtran: Most of what I removed is what I've added in the past. This, along with the other removals, is part of the automated CL2 tests, where we don't need it because it adds extra functionality that isn't being used.  We aren't using report-dir because we aren't keeping track of the CL2 test results, just making sure it's being completed.

eks/cluster-loader/clusterloader2/addon.go
gyuho: Actually, let's skip this (delete). I was confused with other wati func.

gyuho (approved):

PR #182, opened 2020-10-30
Do not set LocationConstraint when creating a S3 bucket in us-east-1

hmizuma:
*Issue #, if available:*

*Description of changes:*

Setting `LocationConstraint` to `us-east-1` when creating a S3 bucket causes `InvalidLocationConstraint`.
This region seems to be handled differently and this param should not be passed in. The location will default to us-east-1.

Reference: https://github.com/boto/boto3/issues/125


gyuho (approved):
lgtm good catch!

PR #183, opened 2020-11-02
 Set us-west-1 to 2 AZ region for VPC creation

hmizuma:
*Issue #, if available:*

*Description of changes:*

This causes issue with AZ index 2 lookup on VPC creation:
`Fn::Select cannot select nonexistent value at index 2`


gyuho (approved):
Lgtm. Thanks

PR #202, opened 2021-02-03
eks: fix list clusters, update CHANGELOG

gyuho:


jayanthvn (approved):
Awesome, thanks for a quick fix :)

PR #334, opened 2023-07-25
Fix broken build; support customization

vincentni:
*Issue #, if available:*
Current build process is broken as it tries to pull some artifacts from upstream s3 buckets, but they are no longer existing.

*Description of changes:*
Fix the broken build process. In addition, add a few build arguments that help customize the built AKT image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,39 +1,53 @@
-FROM golang AS aws-k8s-tester-builder
+ARG GO_VERSION
+ARG AL_VERSION
+
+FROM golang:${GO_VERSION} AS aws-k8s-tester-builder
+ARG GOPROXY
+ARG RELEASE_VERSION
 RUN go version
 ADD ./ /go/src/github.com/aws/aws-k8s-tester
 WORKDIR /go/src/github.com/aws/aws-k8s-tester
-ARG RELEASE_VERSION=latest
-ARG OS_TARGETS=linux
-RUN ./hack/build.sh
+RUN GOPROXY=${GOPROXY} RELEASE_VERSION=${RELEASE_VERSION} ./hack/build.sh
 
-FROM amazonlinux:latest
-RUN yum update -y && yum install -y which python3 pip3 && yum clean all && pip3 install awscli --upgrade --user
+FROM golang:${GO_VERSION} AS clusterloader2-builder
+ARG GOPROXY
+ARG OS_TARGET
+ARG OS_ARCH
+RUN go version
+RUN git clone https://github.com/kubernetes/perf-tests.git /perf-tests
+WORKDIR /perf-tests/clusterloader2
+RUN GOPROXY=${GOPROXY} GOOS=${OS_TARGET} GOARCH=${OS_ARCH} go mod tidy && go build -o ./clusterloader2 ./cmd
+
+FROM amazonlinux:${AL_VERSION}
+ARG K8S_VERSION
+ARG RELEASE_VERSION
+ARG OS_TARGET
+ARG OS_ARCH
+RUN yum update -y && yum install -y which python3 python3-pip && yum clean all && pip3 install awscli --upgrade --user
 ENV PATH=/root/.local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
 RUN echo ${PATH}
-RUN /root/.local/bin/aws --version
 RUN aws --version
 
-ARG RELEASE_VERSION=latest
-COPY --from=aws-k8s-tester-builder /go/src/github.com/aws/aws-k8s-tester/bin/aws-k8s-tester-${RELEASE_VERSION}-linux-amd64 /aws-k8s-tester
-COPY --from=aws-k8s-tester-builder /go/src/github.com/aws/aws-k8s-tester/bin/ec2-utils-${RELEASE_VERSION}-linux-amd64 /ec2-utils
-COPY --from=aws-k8s-tester-builder /go/src/github.com/aws/aws-k8s-tester/bin/ecr-utils-${RELEASE_VERSION}-linux-amd64 /ecr-utils
-COPY --from=aws-k8s-tester-builder /go/src/github.com/aws/aws-k8s-tester/bin/eks-utils-${RELEASE_VERSION}-linux-amd64 /eks-utils
-COPY --from=aws-k8s-tester-builder /go/src/github.com/aws/aws-k8s-tester/bin/etcd-utils-${RELEASE_VERSION}-linux-amd64 /etcd-utils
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -100,10 +104,10 @@ publish: s3-release docker-release
 docker-publish: docker-build docker-push
 
 docker-build:
-	@if [ ! -f "./_tmp/clusterloader2" ]; then echo "downloading clusterloader2"; aws s3 cp --region us-west-2 s3://aws-k8s-tester-public/clusterloader2-linux-amd64 ./_tmp/clusterloader2; else echo "skipping downloading clusterloader2"; fi;
 	@if [ ! -d "${HOME}/go/src/k8s.io/perf-tests" ]; then echo "cloning perf-tests"; mkdir -p ${HOME}/go/src/k8s.io; pushd ${HOME}/go/src/k8s.io; git clone https://github.com/kubernetes/perf-tests.git; popd; else echo "skipping cloning perf-tests"; fi
+	@if [ ! -f "./_tmp/clusterloader2" ]; then echo "compiling clusterloader2"; mkdir _tmp; cd ${HOME}/go/src/k8s.io/perf-tests/clusterloader2; GOPROXY=direct GOOS=${AKT_OS} GOARCH=${AKT_ARCH} go build -o ./clusterloader2 ./cmd; cd -; cp ${HOME}/go/src/k8s.io/perf-tests/clusterloader2/clusterloader2 _tmp; else echo "skipping compiling clusterloader2"; fi;
 	@if [ ! -d "./_tmp/clusterloader2-testing-load" ]; then echo "copying clusterloader2/testing/load"; cp -rf ${HOME}/go/src/k8s.io/perf-tests/clusterloader2/testing/load ./_tmp/clusterloader2-testing-load; else echo "skipping copying clusterloader2/testing/load"; fi
```
cartermckinnon, line 109: Can we move this into the container image? Add another stage to the build and then `COPY` the `clusterloader2` stuff?

Dockerfile
```diff
@@ -1,39 +1,49 @@
...
+ARG OS_ARCH
+
+FROM golang:${GO_VERSION} AS aws-k8s-tester-builder
 RUN go version
 ADD ./ /go/src/github.com/aws/aws-k8s-tester
 WORKDIR /go/src/github.com/aws/aws-k8s-tester
-ARG RELEASE_VERSION=latest
-ARG OS_TARGETS=linux
-RUN ./hack/build.sh
+RUN GOPROXY=direct ./hack/build.sh
```
cartermckinnon, line 11: We should avoid hard-coding `GOPROXY=direct`. Can you pass `GOPROXY` as a `--build-arg` if it's defined?

Makefile
```diff
@@ -100,10 +104,8 @@ publish: s3-release docker-release
 docker-publish: docker-build docker-push
 
 docker-build:
-	@if [ ! -f "./_tmp/clusterloader2" ]; then echo "downloading clusterloader2"; aws s3 cp --region us-west-2 s3://aws-k8s-tester-public/clusterloader2-linux-amd64 ./_tmp/clusterloader2; else echo "skipping downloading clusterloader2"; fi;
-	@if [ ! -d "${HOME}/go/src/k8s.io/perf-tests" ]; then echo "cloning perf-tests"; mkdir -p ${HOME}/go/src/k8s.io; pushd ${HOME}/go/src/k8s.io; git clone https://github.com/kubernetes/perf-tests.git; popd; else echo "skipping cloning perf-tests"; fi
-	@if [ ! -d "./_tmp/clusterloader2-testing-load" ]; then echo "copying clusterloader2/testing/load"; cp -rf ${HOME}/go/src/k8s.io/perf-tests/clusterloader2/testing/load ./_tmp/clusterloader2-testing-load; else echo "skipping copying clusterloader2/testing/load"; fi
-	docker build --network host -t $(AKT_IMG_NAME):$(AKT_TAG) --build-arg RELEASE_VERSION=$(AKT_TAG) .
+	curl -o kubectl -LO https://storage.googleapis.com/kubernetes-release/release/${K8S_VERSION}/bin/${AKT_OS}/${AKT_ARCH}/kubectl
```
cartermckinnon, line 107: could you move this to `hack/build.sh` (since that's what is invoked during the image build)? that way the result can be cached as an image layer versus downloading it every time.

Dockerfile
```diff
@@ -1,39 +1,49 @@
...
-COPY --from=aws-k8s-tester-builder /go/src/github.com/aws/aws-k8s-tester/_tmp/clusterloader2-testing-load/ /
-COPY --from=aws-k8s-tester-builder /go/src/github.com/aws/aws-k8s-tester/_tmp/clusterloader2-testing-load/config.yaml /clusterloader2-test-config.yaml
-COPY --from=aws-k8s-tester-builder /go/src/github.com/aws/aws-k8s-tester/_tmp/clusterloader2 /clusterloader2
+COPY --from=aws-k8s-tester-builder /go/src/github.com/aws/aws-k8s-tester/kubectl /kubectl
+COPY --from=clusterloader2-builder /perf-tests/clusterloader2/testing/load /clusterloader2-testing-load
+COPY --from=clusterloader2-builder /perf-tests/clusterloader2/testing/load/config.yaml /clusterloader2-test-config.yaml
+COPY --from=clusterloader2-builder /perf-tests/clusterloader2/clusterloader2 /clusterloader2
 COPY --from=aws-k8s-tester-builder /go/src/github.com/aws/aws-k8s-tester/eks /eks
 RUN rm -rf /go/src/github.com/aws/aws-k8s-tester
 RUN chmod +x /aws-k8s-tester /cw-utils /ec2-utils /eks-utils /etcd-utils /s3-utils /sts-utils /clusterloader2
```
cartermckinnon, line 43: nit: we can use the `--chmod` flag on each `COPY` to set this, instead of doing it separately

Dockerfile
```diff
@@ -1,39 +1,53 @@
...
 RUN go version
 ADD ./ /go/src/github.com/aws/aws-k8s-tester
 WORKDIR /go/src/github.com/aws/aws-k8s-tester
-ARG RELEASE_VERSION=latest
-ARG OS_TARGETS=linux
-RUN ./hack/build.sh
+RUN GO_PROXY=${GO_PROXY} ./hack/build.sh
+
+FROM golang:${GO_VERSION} AS clusterloader2-builder
+ARG GO_PROXY
```
cartermckinnon, line 12: The `ARG` is an environment variable, so if you call it `GOPROXY` here, you don't need to convert it in every `RUN` (i.e. with `GOPROXY=${GO_PROXY}`).

Similarly, if you pass `--build-arg GOPROXY=${GOPROXY}` in the `Makefile`, you shouldn't need the changes in `hack/build.sh` --`go` will pick up the value from the env.

cartermckinnon (approved):

PR #395, opened 2023-12-04
Adding Network Policy test

jaydeokar:
*Issue #, if available:*
N/A
*Description of changes:*
`go test -v --kubeconfig=$KUBECONFIG`
This adds a basic test for aws-vpc-cni Network Policy. The test does the following - 

* Installs the latest vpc-cni using helm chart and enables network policy support on the cluster
* Creates three servers 
   * Verifies all are reachable when no network policy is applied
   * Verifies traffic is blocked when network policy is applied (which proves the bpf program was loaded)  


e2e2/internal/framework_extensions/helm.go
```diff
@@ -0,0 +1,65 @@
...
+
+	"helm.sh/helm/v3/pkg/action"
+	"helm.sh/helm/v3/pkg/chart/loader"
+	"helm.sh/helm/v3/pkg/cli"
+	"helm.sh/helm/v3/pkg/getter"
+	"helm.sh/helm/v3/pkg/release"
+	"helm.sh/helm/v3/pkg/repo"
+)
+
+func InstallPackagedRelease(repoName string, repoUrl string, chart string, releaseName string, namespace string,
```
cartermckinnon, line 17: Can we reuse any of the helm support from the framework instead of rolling our own? https://github.com/kubernetes-sigs/e2e-framework/blob/c61338a3773c2b64550d0e46037bfe541fa54645/third_party/helm/helm.go

e2e2/internal/framework_extensions/helm.go
```diff
@@ -0,0 +1,65 @@
...
+	client.Namespace = namespace
+	cp, err := client.ChartPathOptions.LocateChart(chart, setting)
+	chartReq, err := loader.Load(cp)
+	release, err := client.Run(chartReq, values)
+	return release, err
+}
+
+func Label(resource string, label string,
+	namespace string, kubeconfig string) ([]byte, error) {
+	cmd := exec.Command("kubectl", "label", "--overwrite", resource,
```
cartermckinnon, line 49: We should use client-go for this instead of calling out to kubectl

e2e2/internal/framework_extensions/helm.go
```diff
@@ -0,0 +1,65 @@
...
+		label, "--namespace", namespace, "--kubeconfig", kubeconfig)
+	return cmd.CombinedOutput()
+}
+
+func ResourceExists(resource string, namespace string, kubeconfig string) ([]byte, error) {
+	cmd := exec.Command("kubectl", "get", resource, "--namespace", namespace,
```
cartermckinnon, line 55: Same here, the e2e-framework has utilities for this

e2e2/test/cases/netpol/main_test.go
```diff
@@ -0,0 +1,333 @@
...
+			for ns, server := range servers {
+				err = createServerAndService(ns, server, 1, client, ctx)
+				if err != nil {
+					log.Printf("Failed to create deployment and service for %s", server)
+					return ctx, err
+				}
+			}
+
+			return ctx, nil
+		},
```
cartermckinnon, line 79: Can you add waiter(s) here so that we don't try to run the test cases until your pre-reqs are met?

e2e2/test/cases/netpol/main_test.go
```diff
@@ -0,0 +1,333 @@
...
+		func(ctx context.Context, config *envconf.Config) (context.Context, error) {
+			client, err := config.NewClient()
+
+			if err != nil {
+				return ctx, err
+			}
+
+			client.Resources().Delete(ctx, &v1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: "a", Namespace: "a"}})
+			client.Resources().Delete(ctx, &v1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: "b", Namespace: "b"}})
+			client.Resources().Delete(ctx, &v1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: "c", Namespace: "c"}})
```
cartermckinnon, line 92: This could be a loop if you moved line 45 up a scope.

e2e2/test/cases/netpol/main_test.go
```diff
@@ -0,0 +1,333 @@
...
+	for _, resource := range resources {
+		if _, err := fwext.ResourceExists(resource, releaseNamespace, config.KubeconfigFile()); err == nil {
+			log.Printf("Patching resource: %s labels and annotations..", resource)
+			fwext.Label(resource, "app.kubernetes.io/managed-by=Helm", releaseNamespace, config.KubeconfigFile())
+			fwext.Annotate(resource, fmt.Sprintf("meta.helm.sh/release-name=%s", releaseName), releaseNamespace, config.KubeconfigFile())
+			fwext.Annotate(resource, fmt.Sprintf("meta.helm.sh/release-namespace=%s", releaseNamespace), releaseNamespace, config.KubeconfigFile())
+		} else {
+			log.Printf("Resource %s does not exist, skipping patching labels/annotations..", resource)
+		}
+	}
```
cartermckinnon, line 120: What's the goal of this?

e2e2/test/cases/netpol/main_test.go
```diff
@@ -0,0 +1,333 @@
...
+
+	cniDS := &appsv1.DaemonSet{
+		ObjectMeta: metav1.ObjectMeta{Name: "aws-node", Namespace: releaseNamespace},
+	}
+
+	wait.For(conditions.New(config.Client().Resources()).ResourceMatch(cniDS, func(object k8s.Object) bool {
+		d := object.(*appsv1.DaemonSet)
+		status := false
+		if d.Status.DesiredNumberScheduled == d.Status.NumberReady {
```
cartermckinnon, line 140: Can you use the existing utility in the framework for your daemonset checks?

https://github.com/kubernetes-sigs/e2e-framework/blob/c61338a3773c2b64550d0e46037bfe541fa54645/klient/wait/conditions/conditions.go#L314

e2e2/test/cases/netpol/main_test.go
```diff
@@ -0,0 +1,333 @@
...
+				t.Fatal("Couldn't connect to server A")
+			}
+			return ctx
+		}).
+		Feature()
+
+	blockCToA := features.New("blockCToA").
+		WithLabel("suite", "netpol").
+		Setup(func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
+			err := fwext.ApplyFiles(cfg.Client().RESTConfig(), manifests)
```
cartermckinnon, line 272: Can you swap this out for `ApplyManifests`? We need to make sure the manifests are embedded in the binary, ex: https://github.com/aws/aws-k8s-tester/blob/main/e2e2/test/cases/nvidia/mpi_test.go#L23-L30

e2e2/internal/framework_extensions/helm.go
jaydeokar: Updated

e2e2/internal/framework_extensions/helm.go
jaydeokar: Updated

e2e2/internal/framework_extensions/helm.go
jaydeokar: Removed this function for now as I'm deleting the resource instead of patching for labels/annotations

e2e2/test/cases/netpol/main_test.go
jaydeokar: The deployment function here uses wait for the resource to be ready
https://github.com/aws/aws-k8s-tester/pull/395/files#diff-4082a58579cf8719172663a0888661788612744805944b7e28a64de150aec95dR226

e2e2/test/cases/netpol/main_test.go
jaydeokar: So vpc-cni is installed by default on a cluster and we need to import the resource into the helm chart (apply annotations/labels).  More details here -
https://github.com/aws/eks-charts/tree/master/stable/aws-vpc-cni#adopting-the-existing-aws-node-resources-in-an-eks-cluster

I've updated this to delete the resources instead of patching for labels/annotations

e2e2/test/cases/netpol/main_test.go
jaydeokar: Seems like we need a e2e-framework release to use this utility. I've added a TODO to update to use this when we have a new version of the framework

e2e2/test/cases/netpol/main_test.go
jaydeokar: I've updated to not use the manifest file and define the network policy using go client

e2e2/test/cases/netpol/main_test.go
cartermckinnon: we could pull in the latest changes with a go `v0.0.0-$COMMIT` version, but you could also use the one I've put in the frameworkextensions package: https://github.com/aws/aws-k8s-tester/blob/main/e2e2/internal/framework_extensions/conditions.go#L32

Dockerfile.kubetest2
```diff
@@ -17,7 +17,7 @@ COPY kubetest2/ .
 RUN go install ./...
 WORKDIR $GOPATH/github.com/aws/aws-k8s-tester/e2e2
 COPY e2e2/ .
-RUN go test -c ./test/cases/nvidia -o $GOPATH/bin/e2e-nvidia
+RUN go test -c ./test/cases/nvidia -o $GOPATH/bin/e2e-nvidia && go test -c ./test/cases/netpol -o $GOPATH/bin/e2e-netpol
```
cartermckinnon, line 20: Can you move this to a separate `RUN` for readability? We're not too concerned about minimizing the number of layers in this image

Dockerfile.kubetest2
```diff
@@ -51,4 +51,5 @@ ENV PATH=$PATH:/info
 RUN cp kubernetes-version.txt /info/
 RUN mv kubernetes/*/bin/* /bin/
 RUN rm -rf /workdir
+RUN curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3 && chmod 700 get_helm.sh && VERIFY_CHECKSUM=false ./get_helm.sh && rm -rf ./get_helm.sh
```
cartermckinnon, line 54: line breaks for readability
```suggestion
RUN curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3 && \
    chmod 700 get_helm.sh && \
    VERIFY_CHECKSUM=false ./get_helm.sh && \
    rm -rf ./get_helm.sh
```

e2e2/test/cases/netpol/main_test.go
```diff
@@ -0,0 +1,425 @@
...
+
+	repoName := "eks"
+	repoUrl := "https://aws.github.io/eks-charts"
+	chartName := "eks/aws-vpc-cni"
+	releaseNamespace := "kube-system"
+
+	client, err := config.NewClient()
+
+	if err != nil {
+		log.Print("Failed to get client")
```
cartermckinnon, line 114: Instead of logging before returning errors throughout this function, you could use `errors.Wrap` if the added context is really valuable

e2e2/test/cases/netpol/main_test.go
```diff
@@ -0,0 +1,425 @@
...
+	client.Resources().Delete(ctx, &rbac.ClusterRoleBinding{ObjectMeta: metav1.ObjectMeta{Name: "aws-node", Namespace: releaseNamespace}})
+
+	// Delete ServiceAccount
+	client.Resources().Delete(ctx, &v1.ServiceAccount{ObjectMeta: metav1.ObjectMeta{Name: "aws-node", Namespace: releaseNamespace}})
+
+	// Delete Daemonset
+	client.Resources().Delete(ctx, &appsv1.DaemonSet{ObjectMeta: metav1.ObjectMeta{Name: "aws-node", Namespace: releaseNamespace}})
+
+	// // Delete Configmap
+	client.Resources().Delete(ctx, &v1.ConfigMap{ObjectMeta: metav1.ObjectMeta{Name: "amazon-vpc-cni", Namespace: releaseNamespace}})
```
cartermckinnon, line 131: This seems fragile to me, we'll have to keep this in sync with the resources that the managed addon for VPC CNI provides

cartermckinnon (changes requested):
My biggest concern is the use of the VPC CNI helm chart, which could change between executions of this test case. I assume you don't plan to use this test case to qualify the network policy agent itself (because it's using the prod helm chart), so it'd be more stable to at least pin a specific version of the chart.

e2e2/test/cases/netpol/main_test.go
jaydeokar: Yes, you are right. I don't think we have an easier way to import the resources into the helm chart at the moment. 
I also considered using eks managed addons to avoid this, but that would mean adding a extra dependency which doesn't seem right here.

I am going to try few more things to see if that helps remove this.
That being said, we have not added new resources to this chart in a while..

jaydeokar: Found a way to not keep this resource mapping by using `helm template` followed by applying the rendered manifest file. However we would need to use another go library for helm client as the inbuilt helm package (e2e-framework) only validates if the chart can be rendered.

cartermckinnon (approved):

cartermckinnon:
Thanks for all the work on this @jaydeokar !

PR #397, opened 2023-12-11
Provision IPv6 clusters via eks api deployer

jaydeokar:
*Issue #, if available:*
N/A

*Description of changes:*
Enable eks api deployer to provision an ipv6 cluster.  By default it falls back to `ipv4` if `--ip-family` option is not  passed

### DIFF
--- a/kubetest2/internal/deployers/eksapi/cloudformation/infra.yaml
+++ b/kubetest2/internal/deployers/eksapi/cloudformation/infra.yaml
@@ -259,7 +259,6 @@ Resources:
 
   #
   # Public subnets
-  # Note: AssignIpv6AddressOnCreation: true is required but a CFN limitation right now, need to add this manually
   SubnetPublic01:
     Type: AWS::EC2::Subnet
     Metadata:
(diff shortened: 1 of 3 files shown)


kubetest2/internal/deployers/eksapi/infra.go
```diff
@@ -85,6 +86,23 @@ func createInfrastructureStack(clients *awsClients, opts *deployerOptions, resou
 		return nil, fmt.Errorf("failed to get infrastructure stack resources: %w", err)
 	}
 	klog.Infof("created infrastructure: %+v", infra)
+
+	if opts.IPFamily == string(ekstypes.IpFamilyIpv6) {
+		for _, subnetID := range infra.subnetsPublic {
+			klog.Infof("enabling auto assign ipv6 on subnet %s", subnetID)
+			_, err = clients._ec2.ModifySubnetAttribute(context.TODO(), &ec2.ModifySubnetAttributeInput{
+				AssignIpv6AddressOnCreation: &ec2types.AttributeBooleanValue{
```
cartermckinnon, line 94: Has CFN still not fixed this? Seems like it's in the schema, does it just not work for public subnets?

https://github.com/aws/aws-k8s-tester/blob/main/kubetest2/internal/deployers/eksapi/cloudformation/infra.yaml#L344

cartermckinnon: You tried adding it to the public subnets in the cloudformation template?

jaydeokar: Good catch. I didn't think of checking if cfn had fixed the issue on their end. 
Update the PR also tested launching a ipv4 and ipv6 cluster, didn't see any issues.

PR #404, opened 2023-12-17
Add metrics registry to kubetest2-eksapi

cartermckinnon:
*Description of changes:*

Adds a metric registry that collects all values and emits then when called. A cloudwatch and no-op implementation are provided.

CloudWatch charges per-PutMetricData call, so the registry is intended to minimize these calls.

Metrics are also added in this PR for the `eksapi` deployer:
- infrasructure stack deletion failures
- leaked ENI's found when deleting infrastructure stack

### DIFF
--- a/kubetest2/internal/deployers/eksapi/infra.go
+++ b/kubetest2/internal/deployers/eksapi/infra.go
@@ -5,17 +5,20 @@ import (
 	_ "embed"
 	"errors"
 	"fmt"
+	"path"
 	"strings"
 	"time"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/service/cloudformation"
 	cloudformationtypes "github.com/aws/aws-sdk-go-v2/service/cloudformation/types"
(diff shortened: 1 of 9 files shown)


kubetest2/internal/deployers/eksapi/infra.go
```diff
@@ -164,7 +183,10 @@ func (m *InfrastructureManager) deleteInfrastructureStack() error {
 			},
 			infraStackDeletionTimeout)
 	if err != nil {
-		return fmt.Errorf("failed to wait for infrastructure stack deletion: %w", err)
+		// don't fail the overall test, the janitor can clean this up
+		klog.Warningf("failed to wait for infrastructure stack deletion: %v", err)
+		m.metrics.Record(infraStackDeletionFailed, 1, nil)
+		return nil
```
mmerkes, line 189: There any risk that this will cause things to look happy and continue our release without us releasing it's failed?

cartermckinnon: Nope, this is just cleanup, it doesn't change anything about the test outcome. The janitor will take care of any leaks, and we don't want to fail the test because of a leaked ENI, for example.

PR #421, opened 2024-02-09
Add addon management to eksapi deployer

cartermckinnon:
*Description of changes:*

Adds a flag (`--addons`) that takes addon name + version pairs (e.g. `vpc-cni:v1.16.0-eksbuild.1`). The addons will be installed before node creation because the addon may be required for the nodes to stabilize.

A specific addon version can be used, or the following "markers":
- `latest`
- `default`

### DIFF
--- a/kubetest2/internal/deployers/eksapi/deployer.go
+++ b/kubetest2/internal/deployers/eksapi/deployer.go
@@ -51,6 +51,7 @@ type deployer struct {
 	metrics          metrics.MetricRegistry
 	infraManager     *InfrastructureManager
 	clusterManager   *ClusterManager
+	addonManager     *AddonManager
 	nodegroupManager *NodegroupManager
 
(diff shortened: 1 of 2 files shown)


kubetest2/internal/deployers/eksapi/deployer.go
```diff
@@ -60,6 +61,7 @@ type deployer struct {
 }
 
 type deployerOptions struct {
+	Addons                      []string      `flag:"addons" desc:"Managed addons (name:version pairs) to create in the cluster. Use 'latest' for the most recent version, or 'default' for the default version."`
```
cartermckinnon, line 64: I wanted to use a `map[string]string` here to get the parsing of the pairs for free, but it didn't work. 🤷

Issacwww (approved):

PR #438, opened 2024-04-10
accept amitype in kubetest2 managed nodegroup creation

ndbaker1:
*Issue #, if available:*

*Description of changes:*

add a flag to pass `ami-type` to the nodegroup creation for kubetest2 deployer

### DIFF
--- a/kubetest2/internal/deployers/eksapi/deployer.go
+++ b/kubetest2/internal/deployers/eksapi/deployer.go
@@ -61,26 +61,27 @@ type deployer struct {
 }
 
 type deployerOptions struct {
-	Addons                      []string      `flag:"addons" desc:"Managed addons (name:version pairs) to create in the cluster. Use 'latest' for the most recent version, or 'default' for the default version."`
-	AMI                         string        `flag:"ami" desc:"AMI for nodes"`
-	ClusterRoleServicePrincipal string        `flag:"cluster-role-service-principal" desc:"Additional service principal that can assume the cluster role"`
-	EFA                         bool          `flag:"efa" desc:"Create EFA interfaces on the node of an unmanaged nodegroup. Requires --unmanaged-nodes."`
-	EKSEndpointURL              string        `flag:"endpoint-url" desc:"Endpoint URL for the EKS API"`
-	EmitMetrics                 bool          `flag:"emit-metrics" desc:"Record and emit metrics to CloudWatch"`
-	ExpectedAMI                 string        `flag:"expected-ami" desc:"Expected AMI of nodes. Up will fail if the actual nodes are not utilizing the expected AMI. Defaults to --ami if defined."`
-	GenerateSSHKey              bool          `flag:"generate-ssh-key" desc:"Generate an SSH key to use for tests. The generated key should not be used in production, as it will not have a passphrase."`
-	InstanceTypes               []string      `flag:"instance-types" desc:"Node instance types"`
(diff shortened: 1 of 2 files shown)


kubetest2/internal/deployers/eksapi/deployer.go
```diff
@@ -263,6 +263,10 @@ func (d *deployer) verifyUpFlags() error {
 	if d.NodeReadyTimeout == 0 {
 		d.NodeReadyTimeout = time.Minute * 5
 	}
+	if d.AMIType == "" {
```
Issacwww, line 266: LGTM, just double check, we won't check against the version to defaulting?

ndbaker1: good point, ill only add the field to the request if the AMIType is provided and non-empty so that it just copies MNG behavior rather than specifying our own

kubetest2/internal/deployers/eksapi/deployer.go
```diff
@@ -61,26 +61,27 @@ type deployer struct {
...
-	Region                      string        `flag:"region" desc:"AWS region for EKS cluster"`
-	TuneVPCCNI                  bool          `flag:"tune-vpc-cni" desc:"Apply tuning parameters to the VPC CNI DaemonSet"`
-	UnmanagedNodes              bool          `flag:"unmanaged-nodes" desc:"Use an AutoScalingGroup instead of an EKS-managed nodegroup."`
-	UpClusterHeaders            []string      `flag:"up-cluster-header" desc:"Additional header to add to eks:CreateCluster requests. Specified in the same format as curl's -H flag."`
-	UserDataFormat              string        `flag:"user-data-format" desc:"Format of the node instance user data"`
+	Addons                      []string          `flag:"addons" desc:"Managed addons (name:version pairs) to create in the cluster. Use 'latest' for the most recent version, or 'default' for the default version."`
+	AMI                         string            `flag:"ami" desc:"AMI for nodes"`
+	AMIType                     ekstypes.AMITypes `flag:"ami-type" desc:"AMI type for nodes"`
```
cartermckinnon, line 66:
```suggestion
	AMIType                     ekstypes.AMITypes `flag:"ami-type" desc:"AMI type for managed nodes"`
```

cartermckinnon:
I think we should add validation that a `--ami-type` and `--ami` aren't used together, the latter only has effect if `--unmaned-nodes` is used.

kubetest2/internal/deployers/eksapi/deployer.go
```diff
@@ -260,6 +260,9 @@ func (d *deployer) verifyUpFlags() error {
 	if d.UnmanagedNodes && d.EFA && len(d.InstanceTypes) != 1 {
 		return fmt.Errorf("--efa requires a single instance type")
 	}
+	if d.AMIType != "" && d.AMI != "" {
+		return fmt.Errorf("Cannot accept mutually exclusive flags --ami-type and --ami because it implies conflicting deployment modes")
```
ndbaker1, line 264: need some suggestion on this message/condition. we don't validate you *can't* provide the `--ami` flag if you're not using unmanaged nodes, so not sure whether enforcing this is useful in practice @cartermckinnon

ndbaker1: validating each on their own according to `--unmanaged-nodes` fixes the mutual exclusion and i think is clearer. small update with this

cartermckinnon (approved):

PR #445, opened 2024-06-06
Add base docker image for EFA NCCL

weicongw:
*Issue #, if available:*

*Description of changes:*
Build a CUDA 12-based Docker image for NCCL testing. The Dockerfile is replicated from https://github.com/aws-samples/aws-efa-nccl-baseami-pipeline/pull/10/files

Issue #, if available:

### DIFF
--- a/kubetest2/internal/deployers/eksapi/nodegroup.go
+++ b/kubetest2/internal/deployers/eksapi/nodegroup.go
@@ -106,7 +106,7 @@ func (m *NodegroupManager) createManagedNodegroup(infra *Infrastructure, cluster
 func (m *NodegroupManager) createUnmanagedNodegroup(infra *Infrastructure, cluster *Cluster, opts *deployerOptions) error {
 	stackName := m.getUnmanagedNodegroupStackName()
 	klog.Infof("creating unmanaged nodegroup stack...")
-	userData, err := generateUserData(opts.UserDataFormat, cluster)
+	userData, userDataIsMimePart, err := generateUserData(opts.UserDataFormat, cluster)
 	if err != nil {
 		return err
 	}
@@ -144,6 +144,10 @@ func (m *NodegroupManager) createUnmanagedNodegroup(infra *Infrastructure, clust
 			{
 				ParameterKey:   aws.String("UserData"),
 				ParameterValue: aws.String(userData),
+			},
+			{
+				ParameterKey:   aws.String("UserDataIsMIMEPart"),
+				ParameterValue: aws.String(strconv.FormatBool(userDataIsMimePart)),
 			},
 			{
 				ParameterKey:   aws.String("ClusterName"),
@@ -203,7 +207,7 @@ func (m *NodegroupManager) createUnmanagedNodegroup(infra *Infrastructure, clust
 func (m *NodegroupManager) createUnmanagedNodegroupWithEFA(infra *Infrastructure, cluster *Cluster, opts *deployerOptions) error {
 	stackName := m.getUnmanagedNodegroupStackName()
 	klog.Infof("creating unmanaged nodegroup with EFA stack...")
-	userData, err := generateUserData(opts.UserDataFormat, cluster)
+	userData, _, err := generateUserData(opts.UserDataFormat, cluster)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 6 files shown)


Dockerfile.nccl
```diff
@@ -0,0 +1,87 @@
+# Start with the NVIDIA CUDA base image
+FROM nvidia/cuda:12.2.0-devel-ubuntu22.04@sha256:7e81c70fb55bfc54e6711700eb721bb94159a3cd546620b37210d12c17914fa6
+
+ARG EFA_INSTALLER_VERSION=latest
+ARG AWS_OFI_NCCL_VERSION=1.7.2
```
Issacwww, line 5: see if there is higher version of this plugin we should bumped up

Dockerfile.nccl
```diff
@@ -0,0 +1,87 @@
+# Start with the NVIDIA CUDA base image
+FROM nvidia/cuda:12.2.0-devel-ubuntu22.04@sha256:7e81c70fb55bfc54e6711700eb721bb94159a3cd546620b37210d12c17914fa6
```
Issacwww, line 2: see if we need higher cuda version 
The sha was added previously because I build it in my local Mac and the arch is not compatible with the test environment. So please check what arch is in the image build env and it would better to remove the sha at the end

Dockerfile.nccl
```diff
@@ -0,0 +1,87 @@
...
+    sed -i 's/#\(StrictModes \).*/\1no/g' /etc/ssh/sshd_config
+ENV LD_LIBRARY_PATH /opt/amazon/openmpi/lib64:/opt/amazon/openmpi/lib:/opt/amazon/efa/lib64:/opt/aws-ofi-nccl/install/lib:/usr/local/cuda/lib64:/usr/local/cuda/extras/CUPTI/lib64:/usr/local/lib/:/usr/lib64:/usr/lib/x86_64-linux-gnu/:$LD_LIBRARY_PATH
+ENV PATH /usr/local/cuda/bin:/opt/amazon/openmpi/bin:/opt/amazon/efa/bin:/usr/sbin:/usr/bin:/usr/local/bin:$PATH
+
+# Install EFA
+RUN cd $HOME \
+    && curl -O https://efa-installer.amazonaws.com/aws-efa-installer-${EFA_INSTALLER_VERSION}.tar.gz \
+    && tar -xf $HOME/aws-efa-installer-${EFA_INSTALLER_VERSION}.tar.gz \
+    && cd aws-efa-installer \
+    && ./efa_installer.sh -y -g -d --skip-kmod --skip-limit-conf --no-verify \
```
Issacwww, line 55: `--no-verify` should be removed, see https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/efa-start.html#efa-start-enable

cartermckinnon: @Issacwww what does that flag do? sounds like it may require an EFA device to be attached or something. We're also skipping the kmod installation, since all we need in the container image is the userspace libraries

weicongw: The doc use the command `sudo ./efa_installer.sh -y` to install efa. Should `-g -d --skip-kmod --skip-limit-conf --no-verify` also be removed?

weicongw: @cartermckinnon detail about the flags
```
 -d, --debug-packages   Install debug packages
 -u, --uninstall        Uninstall EFA packages
 -y, --yes              Run installer without prompting for confirmation
 -n, --no-verify        Skip EFA device verification and test
 -l, --skip-limit-conf  Skip EFA limit configuration
 -k, --skip-kmod        Skip EFA kmod installation
 -g, --enable-gdr       Enable GPUDirect RDMA support. This option is not needed as the support is enabled by default.
 -m, --minimal          Only install kernel module and rdma-core, do not install libfabric and mpi
 --mpi                  A comma-delimited list of MPI flavors(without space) to install. Valid choices are: openmpi4, openmpi5. Only effective if --minimal is not provided.
```

cartermckinnon: >  -n, --no-verify        Skip EFA device verification and test

So, we definitely want to use that flag, because we won't have a real EFA device attached when we're just building the container image. I think all the other flags also make sense. All we need the EFA installer to do is put some libraries on the path in the container image

PR #448, opened 2024-06-12
Remove InstancesDistribution from unmanaged nodegroup cfn

ndbaker1:
*Issue #, if available:*

N/A

*Description of changes:*

Remove `InstancesDistribution` from unmanaged nodegroup cloudformation templates since it may not be supported in certain regions.

**Testing:**

```
# built container
docker build -f Dockerfile.kubetest2 . --build-arg=KUBERNETES_MINOR_VERSION=1.29 -t kubetest2-1.29
docker run --rm -it kubetest2-1.29

# from kubetest2 image context, run test on amazon/amazon-eks-node-al2023-x86_64-standard-1.29-v20240615
AWS_REGION=us-west-2 \
kubetest2 eksapi \
--up \
--down \
--generate-ssh-key \
--expected-ami=ami-03cda10c26c283085 \
--addons=vpc-cni:latest \
--tune-vpc-cni \
--node-name-strategy=EC2PrivateDNSName \
--user-data-format=nodeadm \
--unmanaged-nodes \
--ami=ami-03cda10c26c283085 \
--test=multi \
-- \
--fail-fast=false \
-- \
ginkgo \
--use-binaries-from-path \
--parallel=6 \
--ginkgo-args='--no-color' \
--focus-regex="\\[Conformance\\]" \
--skip-regex="\\[Serial\\]|\\[Disruptive\\]|\\[Slow\\]|Garbage.collector" \
-- \
ginkgo \
--use-binaries-from-path \
--parallel=1 \
--flake-attempts=3 \
--ginkgo-args='--no-color' \
--focus-regex="(\\[Serial\\]|\\[Disruptive\\]|\\[Slow\\]|Garbage.collector).*\\[Conformance\\]" \
--skip-regex=Daemon.set.+should.rollback.without.unnecessary.restarts

...

Ran 40 of 7407 Specs in 2851.458 seconds
SUCCESS! -- 40 Passed | 0 Failed | 0 Pending | 7367 Skipped
PASS

Ginkgo ran 1 suite in 47m32.351613431s
Test Suite Passed
```

### DIFF
--- a/kubetest2/internal/deployers/eksapi/templates/unmanaged-nodegroup-efa.yaml.template
+++ b/kubetest2/internal/deployers/eksapi/templates/unmanaged-nodegroup-efa.yaml.template
@@ -317,8 +317,10 @@ Resources:
     Properties:
       AutoScalingGroupName: !Ref ResourceId
       MixedInstancesPolicy:
+        {{- if .Features.InstancesDistribution }}
         InstancesDistribution:
           OnDemandAllocationStrategy: "prioritized"
+        {{- end }}
         LaunchTemplate:
           LaunchTemplateSpecification:
             LaunchTemplateId: !Ref NodeLaunchTemplate
--- a/kubetest2/internal/deployers/eksapi/deployer.go
+++ b/kubetest2/internal/deployers/eksapi/deployer.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"path"
 	"path/filepath"
-	"strings"
 	"time"
 
 	"github.com/aws/aws-k8s-tester/kubetest2/internal"
@@ -38,11 +37,6 @@ var (
 		Metric:    "TotalRuntimeSeconds",
 		Unit:      cloudwatchtypes.StandardUnitSeconds,
 	}
-)
-
-// default features
-const (
-	InstancesDistribution = "InstancesDistribution"
 )
 
 // assert that deployer implements optional interfaces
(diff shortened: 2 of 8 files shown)


kubetest2/internal/deployers/eksapi/templates/unmanaged-nodegroup-efa.yaml.template
```diff
@@ -317,8 +317,10 @@ Resources:
     Properties:
       AutoScalingGroupName: !Ref ResourceId
       MixedInstancesPolicy:
+        {{- if .Features.InstancesDistribution }}
```
cartermckinnon, line 320: do we even need the `InstancedDistribution`? I just copied this from an unmanaged nodegroup example to begin with, wr can probably just remove it entirely if it's not available everywhere.

*or* we can just use a CFN intrinsic func to switch this off in certain partitions, if that's what we're getting at. Would be easier than pushing that logic up to the user of the deployer IMO

ndbaker1: in that case i think i'd rather pull it out because it doesn't seem required 🤔 

not sure if the parameters will affect runtimes based on capacity of Spot vs OnDemand, but the instance type prioritization seems like the only real different behavior

cartermckinnon: ack, can you test and make sure the nodes still come up without this?

ndbaker1: ran a test successfully and updated the overview with the steps

Issacwww (approved):

cartermckinnon (approved):

PR #455, opened 2024-06-27
Add docker image for BERT e2e inference task

mattcjo:
*Issue #, if available:*

*Description of changes:*
A inference script (e2e2/test/images/bert-inference/infer.py) has been added, along with its dependencies (e2e2/test/images/bert-inference/requirements.txt), in a new docker file (e2e2/test/images/bert-training/Dockerfile.bert-inference). Building the dockerfile will produce an image that will run a BERT inference job on GPU.

The testing of the docker image took place on a `g5.2xlarge` instance utilizing the AMI: ami-05e885690ca33b527. The goal of the image is to run an inference workload through a BERT model. There are two inference modes: `latency` and `throughput`. These are two different modes that optimize for latency or throughput. 

The results of the test show that the running the docker image starts up and executes the BERT inference job for latency optimization as expected :

```
docker run --rm --gpus all -e INFERENCE_MODE=latency aws-bert-inference:latest python infer.py

==========
== CUDA ==
==========

CUDA Version 12.5.0

Container image Copyright (c) 2016-2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.

This container image and its contents are governed by the NVIDIA Deep Learning Container License.
By pulling and using the container, you accept the terms and conditions of this license:
https://developer.nvidia.com/ngc/nvidia-deep-learning-container-license

A copy of this license is made available in this container at /NGC-DL-CONTAINER-LICENSE for your convenience.

/usr/local/lib/python3.11/dist-packages/huggingface_hub/file_download.py:1132: FutureWarning: `resume_download` is deprecated and will be removed in version 1.0.0. Downloads always resume when possible. If you want to force a new download, use `force_download=True`.
  warnings.warn(
GPU is available
Running inference in latency mode with batch size 1
Inference Mode: latency
Average time per batch: 0.0075 seconds
Throughput: 132.61 samples/second
```

The results of the test show that the running the docker image starts up and executes the BERT inference job for throughput optimization as expected :

```
docker run --rm --gpus all -e INFERENCE_MODE=throughput aws-bert-inference:latest python infer.py

==========
== CUDA ==
==========

CUDA Version 12.5.0

Container image Copyright (c) 2016-2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.

This container image and its contents are governed by the NVIDIA Deep Learning Container License.
By pulling and using the container, you accept the terms and conditions of this license:
https://developer.nvidia.com/ngc/nvidia-deep-learning-container-license

A copy of this license is made available in this container at /NGC-DL-CONTAINER-LICENSE for your convenience.

/usr/local/lib/python3.11/dist-packages/huggingface_hub/file_download.py:1132: FutureWarning: `resume_download` is deprecated and will be removed in version 1.0.0. Downloads always resume when possible. If you want to force a new download, use `force_download=True`.
  warnings.warn(
GPU is available
Running inference in throughput mode with batch size 8
Inference Mode: throughput
Average time per batch: 0.0129 seconds
Throughput: 619.77 samples/second
```


weicongw:
I'm not an expert on inference tasks, but can we get the `Sequence Length` in the performance data? Also, how long does each inference take? Can we run multiple inferences and calculate the p50 and p99 latency?

weicongw:
Overall, this PR looks good to me.

mattcjo:
> I'm not an expert on inference tasks, but can we get the `Sequence Length` in the performance data? Also, how long does each inference take? Can we run multiple inferences and calculate the p50 and p99 latency?

So the sequence length will always be 128. As far as p50 vs p99 latency, I'm not sure how much insight we'd gain from these metrics based on how short running of a test it is.

e2e2/test/images/bert-inference/Dockerfile.bert-inference
ndbaker1, line 1: thinking just name this as `Dockerfile`, following the current pattern neuron and nvidia images

mattcjo: I'm good with that. Added on a prefix as I wasn't sure what the directory structure would like at first.

e2e2/test/images/bert-inference/Dockerfile
```diff
@@ -0,0 +1,31 @@
+# Use the NVIDIA CUDA runtime as a parent image
+FROM nvidia/cuda:12.5.0-devel-ubuntu22.04
+
+# Set environment variable to disable interactive prompts
+ENV DEBIAN_FRONTEND=noninteractive
+
+# Install Python 3.11
+RUN apt-get update && apt-get install -y \
+    software-properties-common && \
+    add-apt-repository ppa:deadsnakes/ppa && \
```
ndbaker1, line 10: @mattcjo @weicongw what are your opinions of different ways of installing dependencies like python in these images? I see weicong's is building from source in the neuron image  https://github.com/aws/aws-k8s-tester/blob/cc66356a28d11fdd4a60573d2a2bbe502a14dbab/e2e2/test/images/neuron/Dockerfile#L90-L101

do you see a benefit in being consistent for all our images?

mattcjo: So, in theory I prefer installing directly from source as that gives us the most control and consistency. My main issue with it though is that you need to specify the minor version as well. I feel like using the system package managers do just fine, but would like to hear everyone's take

mattcjo: One other thing to add on... My main point in support of installing from source is that there are sometime new features and optimizations introduced in minor releases. This has the potential of leading to different performance results between minor versions. The likelihood of this happening, or being of any significance? I think pretty low, but something worth noting

ndbaker1: +1 to those points, and irrespective of execution i think having shared expectations across our images is also a good idea (unless there are obvious requirements software/hardware differences)

mattcjo: Sounds good to me. I'll make the changes now

mattcjo: Done

e2e2/test/images/bert-inference/Dockerfile
```diff
@@ -4,20 +4,58 @@ FROM nvidia/cuda:12.5.0-devel-ubuntu22.04
...
-# Create a symbolic link to use python3.11 as python
-RUN ln -sf /usr/bin/python3.11 /usr/bin/python
+RUN apt-get update \
+ && apt-get upgrade -y \
+ && apt-get install -y --no-install-recommends \
+    build-essential \
+    ca-certificates \
+    cmake \
+    curl \
+    emacs \
```
ndbaker1, line 19: if you can trim any of these out I would but nbd 🤔

.github/workflows/ci.yaml
```diff
@@ -30,4 +30,9 @@ jobs:
...
     steps:
     - uses: actions/checkout@v3
-    - run: docker build --file e2e2/test/images/nvidia/Dockerfile .
\ No newline at end of file
+    - run: docker build --file e2e2/test/images/nvidia/Dockerfile .
+  build-bert-inference:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - run: docker build --file e2e2/test/images/bert-inference/Dockerfile e2e2/test/images/bert-inference
```
ndbaker1, line 38: as much as i prefer setting the build workspace to the exact path, I think it's set to `.` in the other images because its built from the root in other workflows. Can you double check this will work

e2e2/test/images/bert-inference/Dockerfile
mattcjo: I actually only included those because they were in our other images - https://github.com/aws/aws-k8s-tester/blob/cc66356a28d11fdd4a60573d2a2bbe502a14dbab/e2e2/test/images/neuron/Dockerfile#L39C1-L43C10

Was just trying to keep things consistent and start to identify common patterns amongst our images. Maybe these are required in the other image...? If not, I would definitely advocate for us to remove.

.github/workflows/ci.yaml
mattcjo: This doesn't work because the image build references `e2e2/test/images/bert-inference/requirements.txt`. If we set the path to `.`, instead of the full path, then it doesn't know it exists. This is actually something I was wanting to discuss, python dependency declaration/management, but was hoping it'd be a down the line thing (not a short convo)

e2e2/test/images/bert-inference/Dockerfile
ndbaker1: yea agreed, its not blocking me

.github/workflows/ci.yaml
ndbaker1: i agree, i don't love the full path but we'll need to make sure it works with our current setup

ndbaker1 (approved):

cartermckinnon (approved):

PR #463, opened 2024-07-31
replace `wait.WithTimeout(timeout)` with `wait.WithContext(ctx))`

weicongw:
*Issue #, if available:*

*Description of changes:*

replace `wait.WithTimeout(timeout)` with `wait.WithContext(ctx))`

### DIFF
--- a/e2e2/test/cases/nvidia/main_test.go
+++ b/e2e2/test/cases/nvidia/main_test.go
@@ -50,6 +50,9 @@ func TestMain(m *testing.M) {
 		log.Fatalf("failed to initialize test environment: %v", err)
 	}
 	testenv = env.NewWithConfig(cfg)
+	ctx, cancel := context.WithTimeout(context.Background(), 60*time.Minute)
+	defer cancel()
+	testenv = testenv.WithContext(ctx)
 
 	// all NVIDIA tests require the device plugin and MPI operator
 	manifests := [][]byte{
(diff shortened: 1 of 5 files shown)


e2e2/test/cases/nvidia/main_test.go
```diff
@@ -50,6 +50,9 @@ func TestMain(m *testing.M) {
 		log.Fatalf("failed to initialize test environment: %v", err)
 	}
 	testenv = env.NewWithConfig(cfg)
+	ctx, cancel := context.WithTimeout(context.Background(), 60*time.Minute)
```
ndbaker1, line 53: this is a timeout for all suites? should we have a timeout per top-level suite like single, multi, etc so that we don't block all the tests if one hangs?

weicongw: Having a timeout per top-level suite is very similar to the the previous approach,.

>>// WithContext provides a way to configure a context that can be used to cancel the wait condition checks. This will enable
// end users to write test in cases where the max timeout is not really predictable or is a factor of a different
// configuration or event.

I think `WithContext` makes more sense, since some test takes longer than others, e.g. nccl test vs unit test

ndbaker1: should use `WithContext` regardless, but i meant using a different context per suite

PR #472, opened 2024-08-30
Fix volume capacity issue

Issacwww:
*Issue #, if available:*
When testing EFA for bottlerocket, run into 
```
 Warning  Evicted              29s   kubelet            The node was low on resource: ephemeral-storage. Threshold quantity: 2867269745, available: 1396568Ki.
 ```

*Description of changes:*
As bottlerocket has 2 volume, we should use `/dev/xvdb` for data volume


ndbaker1 (approved):
sgtm

PR #480, opened 2024-09-16
capacity-resevation requries efa

Issacwww:
*Issue #, if available:*

*Description of changes:*


kubetest2/internal/deployers/eksapi/deployer.go
```diff
@@ -245,6 +245,9 @@ func (d *deployer) verifyUpFlags() error {
 			d.UserDataFormat = "bootstrap.sh"
 			klog.V(2).Infof("Using default user data format: %s", d.UserDataFormat)
 		}
+		if d.CapacityReservation && !d.EFA{
```
bryantbiggs, line 248: not sure if its an implementation detail here, but there is no link between EFA and capacity reservations. There is, however, a requirement for nodes connected via EFA need to be in the same AZ

PR #481, opened 2024-09-16
Fix the AZs when creating subnets

weicongw:
*Issue #, if available:*

*Description of changes:*
This PR addresses an issue with subnet creation by correcting the Availability Zones (AZs) used. Previously, subnets were created in the first two AZs of the region, which often led to failures due to mismatches with capacity reservations. This PR ensures that subnets are created in AZs that align with the specified capacity reservations.


kubetest2/internal/deployers/eksapi/infra.go
```diff
@@ -84,6 +85,31 @@ func (m *InfrastructureManager) createInfrastructureStack(opts *deployerOptions)
...
+			if len(subnetAzs) == 2 {
+				break
+			}
+			if !slices.Contains(subnetAzs, *az.ZoneName) {
+				subnetAzs = append(subnetAzs, *az.ZoneName)
+			}
+		}
+	} else {
+		for i := 0; i < 2; i++ {
+			subnetAzs = append(subnetAzs, *azs.AvailabilityZones[i].ZoneName)
```
ndbaker1, line 109: is the default AZ count guaranteed to be >= 2?

kubetest2/internal/deployers/eksapi/infra.go
```diff
@@ -84,6 +85,31 @@ func (m *InfrastructureManager) createInfrastructureStack(opts *deployerOptions)
...
+		return nil, err
+	}
+	var subnetAzs []string
+	if opts.CapacityReservation {
+		subnetAzs, err = m.getAZsWithCapacity(opts)
+		if err != nil {
+			return nil, err
+		}
+		for _, az := range azs.AvailabilityZones {
+			if len(subnetAzs) == 2 {
```
ndbaker1, line 100: could we have a panic if there is only 1 AZ with reserved capacity?

kubetest2/internal/deployers/eksapi/infra.go
```diff
@@ -15,6 +15,7 @@ import (
 	cloudwatchtypes "github.com/aws/aws-sdk-go-v2/service/cloudwatch/types"
 	"github.com/aws/aws-sdk-go-v2/service/ec2"
 	ec2types "github.com/aws/aws-sdk-go-v2/service/ec2/types"
+	"golang.org/x/exp/slices"
```
ndbaker1, line 18: i thought `slices` was in stdlib 🤔

kubetest2/internal/deployers/eksapi/infra.go
weicongw: We shouldn't, reserved capacity is usually associated with only one AZ.

kubetest2/internal/deployers/eksapi/infra.go
weicongw: I think so, right? Otherwise our original method will also fail.

Issacwww: AFAIK, the CFN we have all create with 2 AZs, and new regions I'm aware of also have 3 AZs by default.

kubetest2/internal/deployers/eksapi/infra.go
Issacwww: I'm ok with this approach, but it would be better if we just create subnets in All AZs and don't worry about reserved capacity's AZ.

Issacwww (approved):

PR #482, opened 2024-09-19
Opt in device plugin

Issacwww:
*Issue #, if available:*

*Description of changes:*
- Make this as flag to support test cases when the device plugin baked in to the AMI
- Remove `[settings.efa]` for BR config as we won't need it after https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/141

### DIFF
--- a/e2e2/test/cases/neuron/main_test.go
+++ b/e2e2/test/cases/neuron/main_test.go
@@ -19,8 +19,9 @@ import (
 )
 
 var (
-	testenv         env.Environment
-	neuronTestImage *string
+	testenv             env.Environment
+	neuronTestImage     *string
+	installDevicePlugin *bool
 )
 
 var (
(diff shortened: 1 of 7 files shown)


e2e2/test/cases/neuron/main_test.go
```diff
@@ -55,6 +58,9 @@ func TestMain(m *testing.M) {
 			return ctx, nil
 		},
 		func(ctx context.Context, config *envconf.Config) (context.Context, error) {
+			if !*installDevicePlugin {
+				return ctx, nil
```
ndbaker1, line 62: feels weird to do this, would be nice if we could add the setup func in the check above but ignore if it makes more trouble

e2e2/test/cases/neuron/main_test.go
```diff
@@ -32,6 +33,7 @@ var (
 
 func TestMain(m *testing.M) {
 	neuronTestImage = flag.String("neuronTestImage", "", "image for neuron single node test")
+	installDevicePlugin = flag.Bool("installDevicePlugin", true, "install nvidia device plugin")
```
ndbaker1, line 36:
```suggestion
	installDevicePlugin = flag.Bool("installDevicePlugin", true, "install neuron device plugin")
```

e2e2/test/cases/neuron/main_test.go
```diff
@@ -30,8 +31,21 @@ var (
 	neuronDevicePluginManifest []byte
 )
 
+func deployNvidiaDevicePlugin(ctx context.Context, config *envconf.Config) (context.Context, error) {
```
ndbaker1, line 34:
```suggestion
func deployNeuronDevicePlugin(ctx context.Context, config *envconf.Config) (context.Context, error) {
```

and should this be with the other helpers?

ndbaker1 (approved):
rest lgtm

PR #486, opened 2024-09-24
Add support to emit metric to the target AMP

weicongw:
Add support to emit metric to the target Amazon Managed Service for Prometheus workspace
Beta

*Issue #, if available:*

*Description of changes:*
- Add support to emit metric to the target Amazon Managed Service for Prometheus workspace
- The test support emitting metric from cross account cross region
- If amp url is not set, the test will not emitting metics
- Emit NCCL test avg bus bandwith metric
- Add metadata label to the metric
- Add/update readme

**Test**
```
go test -timeout 60m -v . -args -nvidiaTestImage public.ecr.aws/o5d5x8n6/weicongw:nvidia --efaEnabled=true --feature=multi-node --ampMetricUrl=https://aps-workspaces.us-west-2.amazonaws.com/workspaces/ws-9f8fe538-f707-46e7-863c-26bfb192dc52/api/v1/remote_write --ampMetricRoleArn=arn:aws:iam::665181186642:role/amp
...
        [1,0]<stdout>:# Out of bounds values : 0 OK
        [1,0]<stdout>:# Avg bus bandwidth    : 3.68456 
        [1,0]<stdout>:#
        [1,0]<stdout>:
        
    mpi_test.go:145: Emitting nccl test metrics to AMP

```
Query the metric from AMP
```
export AMP_QUERY_ENDPOINT=https://aps-workspaces.us-west-2.amazonaws.com/workspaces/ws-9f8fe538-f707-46e7-863c-26bfb192dc52/api/v1/query

awscurl -X POST --region us-west-2 \
--service aps "${AMP_QUERY_ENDPOINT}" \
-d 'query=nccl_average_bandwidth_gbps[60m]' \
--header 'Content-Type: application/x-www-form-urlencoded'

{"status":"success","data":{"resultType":"matrix","result":[{"metric":
{"__name__":"nccl_average_bandwidth_gbps","ami_id":"ami-0cd7612ff47454cd6",
"aws_ofi_nccl_version":"1.9.1","efa_count":"1","efa_enabled":"true",
"efa_installer_version":"1.34.0","instance_type":"p4de.24xlarge",
"kubernetes_version":"1.30+","nccl_version":"2.18.5","node_count":"2",
"nvidia_driver_version":"550.90.07","os_type":"Amazon Linux 2"},
"values":[[1726791286.534,"3.62432"],[1726794564.87,"3.68456"]]}]}}
```


e2e2/internal/utils/aws.go
```diff
@@ -0,0 +1,42 @@
...
+		return aws.Config{}, fmt.Errorf("failed to create AWS SDK config: %v", err)
+	}
+	return c, nil
+}
+
+// creating the payloadHash
+func CreatePayloadHash(req *http.Request) (string, error) {
+	var payloadHash string
+	if req.Body == nil { // For a GET Request (no payload)
+		payloadHash = "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"
```
Issacwww, line 30: should we fail the request for this case?

weicongw: The request will fail eventually. Why do you want to fail the test here?

e2e2/internal/metric/metric.go
```diff
@@ -0,0 +1,192 @@
...
+	return &MetricManager{
+		Metadata:              metadata,
+		AwsConfig:             awsConfig,
+		AmpRemoteWriteURL:     url,
+		AmpRemoteWriteRoleArn: role,
+	}
+}
+
+// PushMetricsToAMP pushes metric data to AWS Managed Prometheus (AMP) using SigV4 authentication
+func (m *MetricManager) PushMetricsToAMP(name string, help string, value float64) error {
```
cartermckinnon, line 43: 1. batching the samples would be preferable to making a separate call to the remote_write API for every sample we collect, IMO

2. are you not able to use the upstream remote write client because of the assume-role jump? https://github.com/prometheus/prometheus/blob/5037cf75f2d4f1671ad365ba1e99902fc36808d5/storage/remote/client.go#L180

e2e2/internal/metric/metric.go
```diff
@@ -0,0 +1,192 @@
...
+	}
+
+	// Snappy compress the Protobuf data
+	compressed := snappy.Encode(nil, data)
+	body := bytes.NewReader(compressed)
+
+	return body, nil
+}
+
+func extractRegionFromURL(remoteWriteURL string) (string, error) {
```
cartermckinnon, line 179: Could you just require the user to pass in the ampRegion? this regex will need to be updated for other+future partitions, etc

e2e2/internal/utils/aws.go
cartermckinnon: this func doesn't seem useful outside `PushMetricsToAmp`, why not just inline it there?

e2e2/test/cases/nvidia/main_test.go
```diff
@@ -129,15 +139,162 @@ func checkNodeTypes(ctx context.Context, config *envconf.Config) (context.Contex
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to list nodes: %v", err)
+	}
+
+	// Ensure at least one node is available
+	if len(nodes.Items) == 0 {
+		return nil, fmt.Errorf("no nodes found in the cluster")
+	}
+
+	// Get instance type and metadata from the first node
```
cartermckinnon, line 172: the test case shouldn't really assume that all the nodes in the cluster are the same across all these dimensions; can you pass in the dimensions with your sample, instead of fetching them ahead of time? Then you'd be able to pass dimensions that you know match the sample

e2e2/test/images/nvidia/Dockerfile
```diff
@@ -97,4 +102,11 @@ RUN rm -rf /var/lib/apt/lists/*
...
-RUN chmod +x ./gpu_unit_tests/unit_test
\ No newline at end of file
+RUN chmod +x ./gpu_unit_tests/unit_test
+
+# Print the values of EFA_INSTALLER_VERSION and NCCL_VERSION
+CMD echo "EFA Installer Version: $EFA_INSTALLER_VERSION" && \
+    echo "NCCL Version: $NCCL_VERSION" && \
+    echo "AWS OFI NCCL Version: $AWS_OFI_NCCL_VERSION" && \
+    printf "NVIDIA Driver Version: " && \
+    nvidia-smi --query-gpu=driver_version --format=csv,noheader | head -n 1
```
cartermckinnon, line 112: I think this would be better suited for an `ENTRYPOINT` script that logged this info and then ran whatever `CMD` was used

e2e2/test/cases/nvidia/main_test.go
```diff
@@ -129,15 +139,162 @@ func checkNodeTypes(ctx context.Context, config *envconf.Config) (context.Contex
...
+		"instance_type":      *nodeType,
+		"node_count":         fmt.Sprintf("%d", nodeCount),
+		"kubernetes_version": k8sVersion,
+		"efa_enabled":        fmt.Sprintf("%t", *efaEnabled),
+		"efa_count":          fmt.Sprintf("%d", efaPerNode),
+		"ami_id":             amiId,
+		"os_type":            osType,
+	}
+
+	// Create a job to fetch the logs of meta info
```
cartermckinnon, line 210: seems like you could just log these details in your actual test run instead of using a separate pod to print them

e2e2/test/cases/nvidia/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+```bash
+docker build --file Dockerfile.kubetest2 --build-arg KUBERNETES_MINOR_VERSION=1.30 --build-arg TARGETOS="linux" --build-arg TARGETARCH="amd64" -t kubetest2 .
+```
+
+### Enter the Kubetest2 Container
+
+```bash
+docker run --name kubetest2 -d -i -t kubetest2 /bin/sh
+docker exec -it kubetest2 sh
```
cartermckinnon, line 30: I would just build the deployer and e2e-nvidia binary locally, would be simpler + faster during dev

e2e2/test/cases/nvidia/main_test.go
weicongw: I couldn't, I tried to add `ENTRYPOINT` in my dockerfile, but the nccl test pods doesn't print these details.

e2e2/test/cases/nvidia/README.md
weicongw: Isn't the `Kubetest2` the deployer?

e2e2/internal/metric/metric.go
weicongw: For the first point, that sounds good—I’ll change it in the next revision. As for the second point, I spent some time trying to use the remote write client, but I wasn’t able to integrate it into my code.

e2e2/internal/utils/aws.go
weicongw: Removed this function and used inline instead.

e2e2/test/cases/nvidia/main_test.go
weicongw: Updated in the latest revision

e2e2/test/images/nvidia/Dockerfile
weicongw: Updated in the latest revision

e2e2/test/cases/nvidia/mpi_test.go
```diff
@@ -160,3 +191,164 @@ func mpiJobSucceeded(obj k8s.Object) bool {
...
+									"nvidia.com/gpu":        node.Status.Capacity["nvidia.com/gpu"],
+									"vpc.amazonaws.com/efa": node.Status.Capacity["vpc.amazonaws.com/efa"],
+								},
+							},
+						},
+					},
+				},
+			},
+		},
+	}
```
Issacwww, line 309: I think we can use a template here to reduce the function size

weicongw: Updated in new rev

e2e2/test/cases/nvidia/mpi_test.go
```diff
@@ -160,3 +199,147 @@ func mpiJobSucceeded(obj k8s.Object) bool {
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to create metadata job: %v", err)
+	}
+
+	// Wait for the job to complete
+	job := &batchv1.Job{
+		ObjectMeta: metav1.ObjectMeta{Name: "metadata-job", Namespace: "default"},
+	}
+	err = wait.For(fwext.NewConditionExtension(cfg.Client().Resources()).JobSucceeded(job),
+		wait.WithContext(ctx))
```
Pavani-Panakanti, line 310: What is this metadata job running here ? Can you also add these details as comments - purpose of this job and what it running

e2e2/test/images/nvidia/Dockerfile
```diff
@@ -1,10 +1,15 @@
 # Start with the NVIDIA CUDA base image
 FROM nvidia/cuda:12.5.0-devel-ubuntu22.04
 
-ARG EFA_INSTALLER_VERSION=latest
+# Add ENV to make ARG values available at runtime
+ARG EFA_INSTALLER_VERSION=1.34.0
+ARG NCCL_VERSION=2.18.5
```
Pavani-Panakanti, line 6: Why are we using an older version of nccl here ? General recommendation is to use either last or last-1 release here

e2e2/internal/metric/metric.go
```diff
@@ -0,0 +1,171 @@
...
+	"github.com/aws/aws-sdk-go-v2/service/sts"
+	"github.com/gogo/protobuf/proto"
+	"github.com/golang/snappy"
+	"github.com/prometheus/client_golang/prometheus"
+	dto "github.com/prometheus/client_model/go"
+	"github.com/prometheus/prometheus/prompb"
+)
+
+type MetricManager struct {
+	// Metadata              map[string]string
```
Pavani-Panakanti, line 26: Are we planning to use this field later ?

Pavani-Panakanti:
@weicongw Can we also rebase the PR ? Thanks

e2e2/test/cases/nvidia/main_test.go
Pavani-Panakanti: Wondering the same..the launcher pods or worker pods should have these details right as they also run the entrypoint script ?

PR #502, opened 2024-11-01
Ensure node for static cluster

Issacwww:
*Issue #, if available:*

*Description of changes:*
For static cluster, use dummy busybox to ensure the node to be ready before e2e test and teardown them after test

## Testing 
```
kubetest2 eksapi --up --down --static-cluster-name=129-static-cluster --test=multi -- --fail-fast=false -- ginkgo --use-binaries-from-path --parallel=6 --ginkgo-args='--no-color' --focus-regex='\[Conformance\]' --skip-regex='\[Serial\]|\[Disruptive\]|\[Slow\]|Garbage.collector'
...
Ran 347 of 7407 Specs in 792.495 seconds
SUCCESS! -- 347 Passed | 0 Failed | 0 Pending | 7060 Skipped

I1101 18:13:18.967304      76 deployer.go:392] Waiting for nodes to be removed: Current node count: 0
```

### DIFF
--- a/kubetest2/internal/deployers/eksapi/deployer.go
+++ b/kubetest2/internal/deployers/eksapi/deployer.go
@@ -1,6 +1,8 @@
 package eksapi
 
 import (
+	"bytes"
+	"context"
 	"flag"
 	"fmt"
 	"path/filepath"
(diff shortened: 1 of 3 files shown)


kubetest2/internal/deployers/eksapi/deployer.go
```diff
@@ -186,6 +194,11 @@ func (d *deployer) Up() error {
 	}
 	if d.deployerOptions.StaticClusterName != "" {
 		klog.Infof("inited k8sclient, skip the rest resource creation for static cluster")
+		if err := d.DeployBusyboxAndWaitForNodes(); err != nil {
```
tzneal, line 197: Why do we need these?

Issacwww: the e2e test suite has a before suite that requires certain nodes to be ready before deploy any testing pods. So deployment this to get nodeclaim work

tzneal (approved):

PR #520, opened 2024-12-13
Create placeholder deployment for Auto clusters if --nodes > 0

cartermckinnon:
*Description of changes:*

This refactors the `NodegroupManager` into a more general `NodeManager` that supports EKS Auto. A "placeholder" deployment will be created that makes the `--nodes` flag effective when `--auto-mode` is used.

There's a similar thing being done for `--static-cluster-name`. That will be removed in later PR's; I haven't touched that behavior so that CI jobs can be migrated without breakage.

### DIFF
--- a/kubetest2/internal/deployers/eksapi/deployer.go
+++ b/kubetest2/internal/deployers/eksapi/deployer.go
@@ -41,7 +41,7 @@ type deployer struct {
 	infraManager         *InfrastructureManager
 	clusterManager       *ClusterManager
 	addonManager         *AddonManager
-	nodegroupManager     *NodegroupManager
+	nodeManager          *nodeManager
 	logManager           *logManager
 	staticClusterManager *StaticClusterManager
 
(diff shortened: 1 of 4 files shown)


kubetest2/internal/deployers/eksapi/deployer.go
```diff
@@ -189,7 +189,7 @@ func (d *deployer) Up() error {
 	if err != nil {
 		return err
 	}
-	if d.deployerOptions.StaticClusterName != "" || d.deployerOptions.AutoMode {
+	if d.deployerOptions.StaticClusterName != "" {
```
cartermckinnon, line 192: This was a bug -- the `staticClusterManager` was likely `nil` if `--auto-mode` was used.

kubetest2/internal/deployers/eksapi/node.go
```diff
@@ -96,7 +105,63 @@ func (m *NodegroupManager) createNodegroup(infra *Infrastructure, cluster *Clust
 	}
 }
 
-func (m *NodegroupManager) createManagedNodegroup(infra *Infrastructure, cluster *Cluster, opts *deployerOptions) error {
+// createPlaceholderDeployment creates a Deployment with the specified number of replicas that requires
+// each replica to be scheduled on different nodes.
+// This ensures that (at least) the specified number of nodes exist in an EKS Auto cluster
+func (m *nodeManager) createPlaceholderDeployment(k8sClient *kubernetes.Clientset, replicas int) (*appsv1.Deployment, error) {
```
tzneal, line 111: Where do these get cleaned up at?

cartermckinnon: I didn't add any cleanup for them, since we can just delete the cluster and this will go down with the ship. Are there any cases that wouldn't work?

tzneal: Ah, yeah that's fine. I was still in static cluster land.

tzneal (approved):

PR #530, opened 2024-12-23
handle the nil pointer exception in writeKubeconfig when no cluster r…

wwvela:
*Issue #, if available:*
* Handle the nil pointer exception in writeKubeconfig method when we didn't set --up flag and --static-cluster-name in kubetest2 command and the cluster pass to writeKubeconfig will be nil.


mattcjo:
LGTM

mattcjo (approved):

PR #536, opened 2025-01-03
Add log collection for --auto-mode

cartermckinnon:
*Description of changes:*

This implements Node log collection when `--auto-mode` is used, using the `NodeDiagnostic` API. Those type definitions aren't in a public package (yet) so i've used the `unstructured` dark arts.


kubetest2/internal/deployers/eksapi/logs.go
```diff
@@ -113,3 +116,130 @@ func (m *logManager) gatherLogsUsingScript(k8sClient *k8sClient, opts *deployerO
...
+		return false, nil
+	})
+	if err != nil {
+		return nil, err
+	}
+	return outcomes, nil
+}
+
+func (m *logManager) isNodeDiagnosticComplete(nodeDiagnostic *unstructured.Unstructured) (bool, []string) {
+	captureStatuses, found, err := unstructured.NestedSlice(nodeDiagnostic.Object, "status", "captureStatuses")
```
ndbaker1, line 219: first time laying eyes on this 😆

ndbaker1 (approved):

kubetest2/internal/deployers/eksapi/logs.go
```diff
@@ -113,3 +116,130 @@ func (m *logManager) gatherLogsUsingScript(k8sClient *k8sClient, opts *deployerO
...
+		if err := k8sClient.client.Create(context.TODO(), &nodeDiagnostic); err != nil {
+			errs = append(errs, err)
+		} else {
+			nodeDiagnostics = append(nodeDiagnostics, nodeDiagnostic)
+		}
+	}
+	outcomes, err := m.waitForNodeDiagnostics(k8sClient, nodeDiagnostics)
+	if err != nil {
+		errs = append(errs, fmt.Errorf("failed to wait for node diagnostics: %v", err))
+	}
```
ndbaker1, line 170: i think we might need to clean up the NodeDiagnostic objects after here. the log collection happens with the `up` and `down` phases, so it would end up blocking the second run and onwards

ndbaker1 (changes requested):

kubetest2/internal/deployers/eksapi/logs.go
cartermckinnon: yup thanks!

kubetest2/internal/deployers/eksapi/logs.go
cartermckinnon: It's better than unboxing it by hand, but still harrowing!

PR #539, opened 2025-01-04
Fix Setup failure: no node has >= 1 GPU(s) in Nvidia Inference test

wwvela:
*Issue #, if available:*
* Nvidia Inference test randomly failed with error `Setup failure: no node has >= 1 GPU(s)`. It is because after apply NVIDIA device plugin it will take some time for the gpu info propagated to the node status. And if the checkGpuCapacity happened before propagation finished, it will failed
* To fix above issue, add retry for GPU capacity check


mattcjo (approved):
LGTM

PR #546, opened 2025-01-07
fix the security group issue

wwvela:
*Issue #, if available:*
* After I going through the control plane logs, I find the csr not approved and a bunch of unverified hosts. And this issue not always happened seems just flacky. I think we bump in this issue that EIP got attached around the same time when the certificate was published.  The issue is happening when customer attaches the EIP to the worker node to Eth0. As a result the public IP address of the node changes from 3.85.202.226 to 3.237.215.201. When API server tries to validate the address, it fails as it doesn't recognize the original public IP address of the node. When kubelet creates a secondary certificate request after 20mins, it succeeds as the new certificate contains EIP address to it. But out test is already finished this time.
* We see this is we use the default cluster security group and attached these netwok interface to it instead of use new EFA security group. And that's why we use newly created EFA security group and connect it with default security group instead of using it directly before.
* To resolve this, we need to roll back the changes to use default cluster security group and handle the eni leakage clean up when deleting the unmanaged node group

### DIFF
--- a/e2e2/test/cases/nvidia-inference/bert_inference_test.go
+++ b/e2e2/test/cases/nvidia-inference/bert_inference_test.go
@@ -71,7 +71,7 @@ func TestBertInference(t *testing.T) {
 			}
 			err := wait.For(
 				fwext.NewConditionExtension(cfg.Client().Resources()).JobSucceeded(job),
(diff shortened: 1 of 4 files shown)


mattcjo (approved):
LGTM

PR #576, opened 2025-01-31
Stabilize neuron training test

mselim00:
*Issue #, if available:*

*Description of changes:*

Neuron training seems to have a NCCL timeout at the first step mark for the first epoch on the very first run on a node, i.e., when the image has to pull. This gets around that by retrying it, since on the retry the image would already be present. The most obvious solution would be using some sort of barrier (e.g. xm.rendezvous() or dist.barrier()) before the first mark step, but those do not seem to accomplish the desired behavior.

### DIFF
--- a/internal/e2e/conditions.go
+++ b/internal/e2e/conditions.go
@@ -3,6 +3,7 @@ package e2e
 import (
 	"context"
 	"fmt"
+	"log"
 
 	appsv1 "k8s.io/api/apps/v1"
 	batchv1 "k8s.io/api/batch/v1"
@@ -77,10 +78,12 @@ func (c *ConditionExtension) ResourceCapacityNonZero(resourceLabel string) apima
         for _, node := range nodeList.Items {
             resource, ok := node.Status.Capacity[v1.ResourceName(resourceLabel)]
             if !ok {
-                return false, fmt.Errorf("Node %s does not have %s capacity defined", node.Name, resourceLabel)
+                log.Printf("Node %s does not have %s capacity defined", node.Name, resourceLabel)
+                return false, nil
             }
             if resource.Value() <= 0 {
-                return false, fmt.Errorf("Node %s has zero %s capacity", node.Name, resourceLabel)
+                log.Printf("Node %s has zero %s capacity", node.Name, resourceLabel)
+                return false, nil
             }
         }
         return true, nil
--- a/internal/e2e/conditions.go
+++ b/internal/e2e/conditions.go
@@ -66,7 +66,7 @@ func (c *ConditionExtension) JobSucceeded(job k8s.Object) apimachinerywait.Condi
 	}
 }
 
-func (c *ConditionExtension) ResourceCapacityNonZero(resourceLabel string) apimachinerywait.ConditionWithContextFunc {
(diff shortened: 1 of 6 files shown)


internal/e2e/conditions.go
```diff
@@ -64,3 +64,25 @@ func (c *ConditionExtension) JobSucceeded(job k8s.Object) apimachinerywait.Condi
...
+        if len(nodeList.Items) == 0 {
+            return false, fmt.Errorf("no nodes found in the cluster")
+        }
+        for _, node := range nodeList.Items {
+            resource, ok := node.Status.Capacity[v1.ResourceName(resourceLabel)]
+            if !ok {
+                return false, fmt.Errorf("Node %s does not have %s capacity defined", node.Name, resourceLabel)
+            }
+            if resource.Value() <= 0 {
+                return false, fmt.Errorf("Node %s has zero %s capacity", node.Name, resourceLabel)
```
mselim00, line 83: errors here would cause a shortcircuit. this seems to be a recoverable state with some wait, we should not error out here so that we can try again with a backoff

internal/e2e/conditions.go
```diff
@@ -48,11 +51,41 @@ func (c *ConditionExtension) JobSucceeded(job k8s.Object) apimachinerywait.Condi
...
+        if err := c.resources.List(ctx, nodeList); err != nil {
+            return false, fmt.Errorf("failed to list nodes: %w", err)
+        }
+        if len(nodeList.Items) == 0 {
+            return false, fmt.Errorf("no nodes found in the cluster")
+        }
+        for _, node := range nodeList.Items {
+            resource, ok := node.Status.Capacity[v1.ResourceName(resourceLabel)]
+            if !ok {
+                log.Printf("Node %s does not have %s capacity defined", node.Name, resourceLabel)
```
mselim00, line 81: prob want to avoid logging in a util func

test/cases/neuron-training/main_test.go
```diff
@@ -184,3 +184,46 @@ func checkNodeTypes(ctx context.Context, config *envconf.Config) (context.Contex
...
+    
+    // Check Neuron devices
+    log.Println("Checking Neuron device capacity on nodes")
+    err := wait.For(
+        fwext.NewConditionExtension(config.Client().Resources()).ResourceCapacityNonZero("aws.amazon.com/neuron"),
+        wait.WithTimeout(time.Second*60),
+        wait.WithInterval(time.Second*5),
+    )
+    if err != nil {
+        return ctx, fmt.Errorf("timeout waiting for Neuron devices to be available: %w", err)
```
mselim00, line 199: should be more generic, could also error for no nodes being found

internal/e2e/conditions.go
```diff
@@ -48,11 +51,41 @@ func (c *ConditionExtension) JobSucceeded(job k8s.Object) apimachinerywait.Condi
...
+			}
+		}
 		if status.Succeeded != *spec.Completions {
 			return false, nil
 		}
 		return true, nil
 	}
 }
+
+func (c *ConditionExtension) ResourceCapacityNonZero(resourceLabel string) apimachinerywait.ConditionWithContextFunc {
```
mselim00, line 69: try to reflect that this checks for all nodes, rather than cluster level nonzero availability

wwvela (approved):
LGTM

PR #577, opened 2025-01-31
fix(neuron/multi-node): Wait for neuron device info to propagate, enable run on more than 2 nodes

mselim00:
*Issue #, if available:*

*Description of changes:*

Some changes so this test can support >2 nodes, and avoid flakes caused by early reads of neuron device information from the node before values are set.

Ran on 4 trn1.32xlarge nodes:
```
$ go test ./... -v -tags=e2e -timeout=60m --test.run=^TestNeuronNodes$/multi-node --efaEnabled=true --neuronTestImage=<IMAGE_URI> --nodeType=trn1.32xlarge 

$ k get pods
NAME                                   READY   STATUS    RESTARTS        AGE
multi-node-nccom-test-launcher-2z5lz   1/1     Running   2 (3m33s ago)   3m36s
multi-node-nccom-test-worker-0         1/1     Running   0               3m36s
multi-node-nccom-test-worker-1         1/1     Running   0               3m36s
multi-node-nccom-test-worker-2         1/1     Running   0               3m36s
multi-node-nccom-test-worker-3         1/1     Running   0               3m36s

[output]
... 
        [1,0]<stdout>:+---+----+[1,0]<stdout>:---------+---------+------------+-------+--------+[1,0]<stdout>:----------+----------+----------+--------+---------+---------+-------+
               size(B)    count(elems)    type    time:avg(us)    algbw(GB/s)    busbw(GB/s)
                     8               2    fp32         1209.66           0.00           0.00
                    16               4    fp32         1212.37           0.00           0.00
                    32               8    fp32         1195.08           0.00           0.00
                    64              16    fp32         1065.07           0.00           0.00
                   128              32    fp32         1200.62           0.00           0.00
                   256              64    fp32         1215.22           0.00           0.00
                   512             128    fp32         1219.41           0.00           0.00
                  1024             256    fp32         1230.95           0.00           0.00
                  2048             512    fp32         1232.07           0.00           0.00
                  4096            1024    fp32          1237.5           0.00           0.01
                  8192            2048    fp32         1226.38           0.01           0.01
                 16384            4096    fp32         1223.28           0.01           0.03
                 32768            8192    fp32         1221.73           0.03           0.05
                 65536           16384    fp32         1234.03           0.05           0.11
                131072           32768    fp32         1229.32           0.11           0.21
                262144           65536    fp32         1234.87           0.21           0.42
                524288          131072    fp32         1245.72           0.42           0.84
               1048576          262144    fp32         1440.84           0.73           1.44
               2097152          524288    fp32         1425.63           1.47           2.92
               4194304         1048576    fp32          682.31           6.15          12.20
               8388608         2097152    fp32          637.75          13.15          26.10
              16777216         4194304    fp32         1026.31          16.35          32.44
              33554432         8388608    fp32          1239.4          27.07          53.72
              67108864        16777216    fp32         2129.18          31.52          62.54
             134217728        33554432    fp32         4342.88          30.91          61.33
             268435456        67108864    fp32         8237.32          32.59          64.67
             536870912       134217728    fp32        13939.16          38.52          76.43
            1073741824       268435456    fp32        22432.64          47.87          94.98
            2147483648       536870912    fp32        44185.72          48.60          96.44
        Avg bus bandwidth:	20.2378GB/s

```

### DIFF
--- a/internal/e2e/resources.go
+++ b/internal/e2e/resources.go
@@ -6,13 +6,13 @@ import (
 	v1 "k8s.io/api/core/v1"
 )
 
-func GetNonZeroResourceCapacityOrError(node v1.Node, resourceName string) (int, error) {
+func GetNonZeroResourceCapacity(node v1.Node, resourceName string) (int, error) {
 	capacity, ok := node.Status.Capacity[v1.ResourceName(resourceName)]
 	if !ok {
-		return 0, fmt.Errorf("node \"%s\" has no resource \"%s\"", node.Name, resourceName)
+		return 0, fmt.Errorf("node %q has no resource %q", node.Name, resourceName)
 	}
 	if capacity.Value() == 0 {
-		return 0, fmt.Errorf("node \"%s\" has zero capacity for resource \"%s\"", node.Name, resourceName)
+		return 0, fmt.Errorf("node %q has zero capacity for resource %q", node.Name, resourceName)
 	}
 	return int(capacity.Value()), nil
 }
--- a/internal/e2e/resources.go
+++ b/internal/e2e/resources.go
@@ -6,7 +6,7 @@ import (
 	v1 "k8s.io/api/core/v1"
 )
 
-func GetNonZeroResourceCapacity(node v1.Node, resourceName string) (int, error) {
+func GetNonZeroResourceCapacity(node *v1.Node, resourceName string) (int, error) {
 	capacity, ok := node.Status.Capacity[v1.ResourceName(resourceName)]
 	if !ok {
 		return 0, fmt.Errorf("node %q has no resource %q", node.Name, resourceName)
--- a/internal/e2e/resources.go
+++ b/internal/e2e/resources.go
@@ -1,0 +1,18 @@
+package e2e
+
+import (
+	"fmt"
+
+	v1 "k8s.io/api/core/v1"
+)
+
+func GetNonZeroResourceCapacityOrError(node v1.Node, resourceName string) (int, error) {
+	capacity, ok := node.Status.Capacity[v1.ResourceName(resourceName)]
+	if !ok {
+		return 0, fmt.Errorf("node \"%s\" has no resource \"%s\"", node.Name, resourceName)
+	}
+	if capacity.Value() == 0 {
+		return 0, fmt.Errorf("node \"%s\" has zero capacity for resource \"%s\"", node.Name, resourceName)
+	}
+	return int(capacity.Value()), nil
+}
--- a/test/cases/neuron/main_test.go
+++ b/test/cases/neuron/main_test.go
@@ -108,7 +108,7 @@ func checkNodeTypes(ctx context.Context, config *envconf.Config) (context.Contex
 		return ctx, fmt.Errorf("no nodes found in the cluster")
 	}
 
-	var nodeCount, totalEfaCount, totalNeuronCoreCount, totalNeuronCount int
+	var totalEfaCount, totalNeuronCoreCount, totalNeuronCount int
 	if *nodeType == "" {
 		nodeType = aws.String(nodes.Items[0].Labels["node.kubernetes.io/instance-type"])
 		log.Printf("No node type specified. Using the node type %s in the node groups.", *nodeType)
--- a/test/cases/neuron/main_test.go
+++ b/test/cases/neuron/main_test.go
@@ -115,14 +115,14 @@ func checkNodeTypes(ctx context.Context, config *envconf.Config) (context.Contex
 	}
 	for _, node := range nodes.Items {
 		if node.Labels["node.kubernetes.io/instance-type"] == *nodeType {
-			neuron, err := e2e.GetNonZeroResourceCapacityOrError(node, "aws.amazon.com/neuron")
+			neuron, err := e2e.GetNonZeroResourceCapacity(&node, "aws.amazon.com/neuron")
 			if err != nil {
 				return nil, err
 			}
 			totalNeuronCount += neuron
 
 			// Check for NeuronCore capacity
-			neuronCore, err := e2e.GetNonZeroResourceCapacityOrError(node, "aws.amazon.com/neuroncore")
+			neuronCore, err := e2e.GetNonZeroResourceCapacity(&node, "aws.amazon.com/neuroncore")
 			if err != nil {
 				return nil, err
 			}
@@ -130,7 +130,7 @@ func checkNodeTypes(ctx context.Context, config *envconf.Config) (context.Contex
 
 			// Check for EFA capacity
 			if *efaEnabled {
-				efa, err := e2e.GetNonZeroResourceCapacityOrError(node, "vpc.amazonaws.com/efa")
+				efa, err := e2e.GetNonZeroResourceCapacity(&node, "vpc.amazonaws.com/efa")
 				if err != nil {
 					return nil, err
 				}
--- a/test/cases/neuron/main_test.go
+++ b/test/cases/neuron/main_test.go
@@ -13,7 +13,9 @@ import (
 	"testing"
 	"time"
 
+	"github.com/aws/aws-k8s-tester/internal/e2e"
 	fwext "github.com/aws/aws-k8s-tester/internal/e2e"
+	"github.com/aws/aws-sdk-go-v2/aws"
 	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -90,6 +92,8 @@ func deployEFAPlugin(ctx context.Context, config *envconf.Config) (context.Conte
 }
 
 func checkNodeTypes(ctx context.Context, config *envconf.Config) (context.Context, error) {
+	time.Sleep(time.Minute) // give node info time to populate
+
 	clientset, err := kubernetes.NewForConfig(config.Client().RESTConfig())
 	if err != nil {
 		return ctx, fmt.Errorf("failed to create Kubernetes client: %w", err)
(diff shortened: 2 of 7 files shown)


test/cases/neuron/manifests/multi-node-test-neuron.yaml
```diff
@@ -24,82 +24,35 @@ spec:
...
-
-                # Read the IPs from the hosts.txt file
-                WORKER_0_IP=$(sed -n '1p' /mnt/data/hosts.txt | tr -d '\r' | tr -d '\n' )
-                WORKER_1_IP=$(sed -n '2p' /mnt/data/hosts.txt | tr -d '\r' | tr -d '\n' )
+            - | 
+                WORKER_0_IP=$(getent hosts multi-node-nccom-test-worker-0.multi-node-nccom-test | awk '{print $1}')
+                WORKER_IPS=$WORKER_0_IP
+                for i in $(seq 1 $(({{.WorkerNodeCount}} - 1))); do
+                  WORKER_IPS="$WORKER_IPS $(getent hosts multi-node-nccom-test-worker-$i.multi-node-nccom-test | awk '{print $1}')"
+                done
```
ndbaker1, line 34: could we just combine these?

```suggestion
                for i in $(seq 0 $(({{.WorkerNodeCount}} - 1))); do
                  WORKER_IPS="$WORKER_IPS $(getent hosts multi-node-nccom-test-worker-$i.multi-node-nccom-test | awk '{print $1}')"
                done
```

test/cases/neuron/manifests/multi-node-test-neuron.yaml
```diff
@@ -24,82 +24,35 @@ spec:
...
 
                 export CCOM_SOCKET_IFNAME=eth0
                 export NEURON_RT_ROOT_COMM_ID=$WORKER_0_IP:63182
+                export OMPI_ALLOW_RUN_AS_ROOT=1
+                export OMPI_ALLOW_RUN_AS_ROOT_CONFIRM=1
```
ndbaker1, line 39: can you summarize here or in the overview what the issues are with running non-root? we should propose the options before committing to this

test/cases/neuron/neuron_test.go
```diff
@@ -70,7 +71,9 @@ func TestNeuronNodes(t *testing.T) {
 			}
 			t.Log("Waiting for single node job to complete")
 			err := wait.For(fwext.NewConditionExtension(cfg.Client().Resources()).JobSucceeded(job),
-				wait.WithContext(ctx))
+				wait.WithContext(ctx),
+				wait.WithTimeout(time.Minute*20),
```
ndbaker1, line 75: thanks for these explicit timeouts 👍

test/cases/neuron/main_test.go
```diff
@@ -203,6 +202,7 @@ func TestMain(m *testing.M) {
 		setUpFunctions = append(setUpFunctions, deployEFAPlugin)
 	}
 
+	setUpFunctions = append(setUpFunctions, checkNonZeroResourceCapacity, checkNodeTypes)
```
ndbaker1, line 205: nice catch!

ndbaker1 (commented):
nice improvements, let talk about the root container changes before continuing

test/cases/neuron/manifests/multi-node-test-neuron.yaml
mselim00: The main issue running the launcher container as non-root in this case is ssh is just trickier to set up. The worker pods **must** run sshd as root. The previous approach was to run all containers as non-root and set up password for the worker pods to use to assume root, start the daemon, and then infinitely sleep in a loop. Switching to root here avoids that extra setup.

There shouldn't actually be any harm using root here and it helps create for a more concise/readable job spec. 

For reference, OpenMPI discourages running as root because:
> Running as root exposes the user to potentially catastrophic file system corruption and damage — e.g., if the user accidentally points the root of the session directory to a system required point, this directory and all underlying elements will be deleted upon job completion, thereby rendering the system inoperable.
https://docs.open-mpi.org/en/main/man-openmpi/man1/mpirun.1.html#the-allow-run-as-root-option

This is ran in a container with no host mount, and probably also on an ephemeral test host, so the blast radius here is really small.


That said, I don't think we should keep this as root in the long-run just out of principal. This, and all the other tests (e.g. [nvidia training](https://github.com/aws/aws-k8s-tester/blob/main/test/cases/nvidia-training/manifests/bert-training.yaml#L29)) should eventually be refactored so the launcher is non-root.

test/cases/neuron/manifests/multi-node-test-neuron.yaml
mselim00: 👍 sounds good
one thing here was we need the first worker to set an env var below
```
export NEURON_RT_ROOT_COMM_ID=$WORKER_0_IP:63182
```
but we can just make this an actual array and index it

test/cases/neuron/main_test.go
```diff
@@ -222,3 +222,48 @@ func TestMain(m *testing.M) {
...
 	os.Exit(testenv.Run(m))
 }
+
+func checkNonZeroResourceCapacity(ctx context.Context, config *envconf.Config) (context.Context, error) {
+	log.Println("[INFO] Starting resource capacity checks")
+
+	// Check Neuron devices
+	log.Println("Checking Neuron device capacity on nodes")
+	err := wait.For(
+		fwext.NewConditionExtension(config.Client().Resources()).AllNodesHaveNonZeroResourceCapacity("aws.amazon.com/neuron"),
```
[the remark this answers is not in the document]
mselim00: agreed, will make that change in a future PR because neuron-training (and maybe others) also makes this assumption

test/cases/neuron/manifests/multi-node-test-neuron.yaml
```diff
@@ -24,82 +24,35 @@ spec:
...
             name: nccom-test-worker
-            command: ["/bin/sh", "-c"]
-            args: ["echo password | sudo -S service ssh start; while true; do sleep 30; done"]
             imagePullPolicy: Always
-            securityContext:
-              privileged: false
             resources:
               limits:
                 aws.amazon.com/neuron: {{.NeuronPerNode}}
+                aws.amazon.com/neuroncore: {{.NeuronCorePerNode}}
```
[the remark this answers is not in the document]
mselim00: nothing at the moment, but this could be an issue in the future. This test utilizes all the neuron cores on the node but doesn't explicitly request it, so if another test just requested the neuron core and not `neuron`,  they would conflict

test/cases/neuron/manifests/multi-node-test-neuron.yaml
mselim00: Did this in [ec208b7](https://github.com/aws/aws-k8s-tester/pull/577/commits/ec208b7b9bceb55aae012bf5bba5ae33823e9ac7)

test/cases/neuron/main_test.go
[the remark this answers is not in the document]
mselim00: I worked this in here, lmk what you think. Switched it for a 1 min sleep and then errored if any expected resources are still zero for just the nodeType set or assumed

internal/e2e/resources.go
```diff
@@ -0,0 +1,18 @@
...
+	v1 "k8s.io/api/core/v1"
+)
+
+func GetNonZeroResourceCapacityOrError(node v1.Node, resourceName string) (int, error) {
+	capacity, ok := node.Status.Capacity[v1.ResourceName(resourceName)]
+	if !ok {
+		return 0, fmt.Errorf("node \"%s\" has no resource \"%s\"", node.Name, resourceName)
+	}
+	if capacity.Value() == 0 {
+		return 0, fmt.Errorf("node \"%s\" has zero capacity for resource \"%s\"", node.Name, resourceName)
```
ndbaker1, line 15:
```suggestion
		return 0, fmt.Errorf("node %q has zero capacity for resource %q", node.Name, resourceName)
```

internal/e2e/resources.go
```diff
@@ -0,0 +1,18 @@
...
+import (
+	"fmt"
+
+	v1 "k8s.io/api/core/v1"
+)
+
+func GetNonZeroResourceCapacityOrError(node v1.Node, resourceName string) (int, error) {
+	capacity, ok := node.Status.Capacity[v1.ResourceName(resourceName)]
+	if !ok {
+		return 0, fmt.Errorf("node \"%s\" has no resource \"%s\"", node.Name, resourceName)
```
ndbaker1, line 12:
```suggestion
		return 0, fmt.Errorf("node %q has no resource %q", node.Name, resourceName)
```

internal/e2e/resources.go
```diff
@@ -0,0 +1,18 @@
+package e2e
+
+import (
+	"fmt"
+
+	v1 "k8s.io/api/core/v1"
+)
+
+func GetNonZeroResourceCapacityOrError(node v1.Node, resourceName string) (int, error) {
```
ndbaker1, line 9: naming is redundant when the types imply the behavior. `OrDie` works if theres a `panic` but that's hidden unless you read the implementation
```suggestion
func GetNonZeroResourceCapacity(node v1.Node, resourceName string) (int, error) {
```

test/cases/neuron/main_test.go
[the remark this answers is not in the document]
ndbaker1: im alright with this, since it's still a strictly better improvement

mselim00:
Confirmed the test still works by re-runnning
```
go test -v -tags=e2e --test.run TestNeuronNodes/multi-node --nodeType=trn1n.32xlarge  --neuronTestImage=<IMAGE_URI> --efaEnabled=true --test.timeout=60m
```

test/cases/neuron/main_test.go
```diff
@@ -104,44 +108,35 @@ func checkNodeTypes(ctx context.Context, config *envconf.Config) (context.Contex
...
 	}
 
-	// Check if all nodes have the same instance type
-	for i := 1; i < len(nodes.Items); i++ {
-		if nodes.Items[i].Labels["node.kubernetes.io/instance-type"] != nodes.Items[i-1].Labels["node.kubernetes.io/instance-type"] {
-			return ctx, fmt.Errorf("inconsistent node types detected, all nodes must have the same instance type")
-		}
+	var totalEfaCount, totalNeuronCoreCount, totalNeuronCount int
+	if *nodeType == "" {
+		nodeType = aws.String(nodes.Items[0].Labels["node.kubernetes.io/instance-type"])
```
ndbaker1, line 113: do we at least want to try finding a suitable node type? since anything non-neuron will fail the test

test/cases/neuron/main_test.go
```diff
@@ -104,44 +108,35 @@ func checkNodeTypes(ctx context.Context, config *envconf.Config) (context.Contex
...
-
-		// Check for NeuronCore capacity
-		neuronCore, ok := node.Status.Capacity["aws.amazon.com/neuroncore"]
-		if ok {
-			totalNeuronCoreCount += int(neuronCore.Value())
-		} else {
-			log.Printf("[WARN] Node %s does not have 'aws.amazon.com/neuroncore' capacity", node.Name)
-		}
+		if node.Labels["node.kubernetes.io/instance-type"] == *nodeType {
+			neuron, err := e2e.GetNonZeroResourceCapacity(&node, "aws.amazon.com/neuron")
```
ndbaker1, line 118: nit: i like to flatten code whenever possible, i.e. making this `continue` if `!= nodeType`. up to you though

ndbaker1 (approved):
last few nits but lgtm

test/cases/neuron/main_test.go
mselim00: It can get pretty convoluted trying to infer it when there's mixed types, particularly if there's two different types with `neuron` capacity. This is just a little bit more gentle than fast failing but makes it so the test will run successfully if there's a single node type in the cluster, but makes it obvious in other cases by printing both the nodeType inferred and showing the error from the first node not having `neuron` capacity.

mselim00:
Images are failing to build on a mongo driver lookup
>  461.7 ../../../../pkg/mod/github.com/go-openapi/strfmt@v0.23.0/bson.go:24:2: unrecognized import path "go.mongodb.org/mongo-driver": https fetch: Get "https://go.mongodb.org/mongo-driver?go-get=1": dial tcp: lookup go.mongodb.org on 168.63.129.16:53: no such host

I can build the image locally and I've retried it a couple times here - seems to be some weird caching causing this failure to persist

ndbaker1 (approved):

PR #588, opened 2025-02-21
fix: try to find suitable AZ for EFA, honor capacity-reservation for non-EFA

mselim00:
*Issue #, if available:*

*Description of changes:*

This is primarily intended to fix two things:
1. `--capacity-reservation` flag was previously ignored when not combined with `--efa`
2. Instance launches with EFA were frequently ICE'd, in part because we sometimes tried to launch in an availability zone that does not offer that instance type. This improves that by checking the availability zones in the VPC for that availability zone, but there's currently only two, so it's possible neither will have that instance type offered. At least in that case, we can fail the doomed test run a little bit faster.

Also a small fix - EFA nodegroups previously used the infraStackCreationTimeout rather than the nodegroup creation timeout. 

For 2, it's worth noting that per the [DescribeAvailabilityZones API docs](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeAvailabilityZones.html), the order of the output is not fixed. So when we select the two first availability zones, we're not necessarily getting `a` and `b`. This means we can't rely on any theoretical observations like `a` tends to have all the instance types offered, and instead we should just be creating a subnet in every AZ to guarantee we have one included. I'll do this in another PR because Cloudformation doesn't natively support that, so it'll get a bit hairy switching things over to a go template.

The other main change in this PR is a combination of the unmanaged nodegroup stacks (EFA and non-EFA). This will help maintenance going forward and to avoid oversights like (1) with the capacity reservation flag being selectively honored. The main incompatibilities between the two stacks are the subnet and network interface handling, but those should be resolved with this PR, making the stack creation more extensible for other features we want to add in the future without duplicating the work for two stacks.

Anyone calling this deployer with `--efa` without the `--capacity-reservation` flag will need to make sure the role used to deploy it has `ec2:DescribeInstanceTypeOfferings` permissions.


internal/deployers/eksapi/node.go
```diff
@@ -158,6 +154,11 @@ func (m *nodeManager) resolveInstanceTypes(opts *deployerOptions) (err error) {
 	if len(validInstanceTypes) == 0 {
 		return fmt.Errorf("none of the instance types %v were valid", instanceTypes)
 	}
+	if opts.EFA {
+		// EFA instances require their own launch template, so only one can be used
+		klog.Info("using just the first instance type because EFA was requested")
+		validInstanceTypes = []string{validInstanceTypes[0]}
```
ndbaker1, line 160: we should fail here, we can't just hijack the instance types IMO

internal/deployers/eksapi/templates/unmanaged-nodegroup.yaml.template
```diff
@@ -42,60 +41,120 @@ Parameters:
...
+                  SubnetId: {{$networkInterface.SubnetId}}
+                  DeleteOnTermination: {{$networkInterface.DeleteOnTermination}}
+                {{- end}}
+            - !Ref AWS::NoValue
         SecurityGroupIds:
-          - !Ref SecurityGroup
+            Fn::If:
+            - IsEFAEnabled
+            - !Ref AWS::NoValue
+            - - !Ref SecurityGroup
```
ndbaker1, line 157: why dont we just always pass this to avoid confusion?

internal/deployers/eksapi/templates/unmanaged-nodegroup.yaml.template
```diff
@@ -120,15 +179,32 @@ Resources:
...
+              {{- end}}
+      DesiredCapacity: !Ref NodeCount
+      MinSize: !Ref NodeCount
+      MaxSize: !Ref NodeCount
+      VPCZoneIdentifier: !Ref SubnetIds
+      Tags:
+        - Key: Name
+          Value: !Sub "${ClusterName}-Node"
+          PropagateAtLaunch: true
+        - Key: !Sub "kubernetes.io/cluster/${ClusterName}"
```
ndbaker1, line 208: saw this was copied from above and figure we should just keep the comment for context.
```suggestion
        # necessary for kubelet's legacy, in-tree cloud provider
        - Key: !Sub "kubernetes.io/cluster/${ClusterName}"
```

internal/deployers/eksapi/templates/unmanaged-nodegroup.yaml.template
```diff
@@ -42,60 +41,120 @@ Parameters:
...
+        CapacityReservationSpecification:
+          Fn::If:
+            - IsCapacityReservationIdSet
+            - CapacityReservationTarget:
+                CapacityReservationId: !Ref CapacityReservationId
+            - !Ref AWS::NoValue
+        IamInstanceProfile:
+          Arn: !GetAtt NodeInstanceProfile.Arn
+        ImageId: !Ref AMIId
+        InstanceType: !Ref InstanceType
```
ndbaker1, line 137: did you need to introduce this in the cfn and pass it from go `instanceTypes[0]` when you could keep using `"{{index .InstanceTypes 0}}"`?

internal/deployers/eksapi/templates/unmanaged-nodegroup.yaml.template
```diff
@@ -42,60 +41,120 @@ Parameters:
...
+            Ebs:
+              DeleteOnTermination: true
+              VolumeSize: !Ref NodeDiskSize
+              VolumeType: gp2
+        CapacityReservationSpecification:
+          Fn::If:
+            - IsCapacityReservationIdSet
+            - CapacityReservationTarget:
+                CapacityReservationId: !Ref CapacityReservationId
+            - !Ref AWS::NoValue
```
ndbaker1, line 133: neat, so capacity reservations can be used in any unmanaged nodegroup now

ndbaker1 (commented):
happy to be merging templates, left some questions

internal/deployers/eksapi/templates/unmanaged-nodegroup.yaml.template
mselim00: Unfortunately this is enforced by launch template creation
> SecurityGroupIds
 The IDs of the security groups. You can specify the IDs of existing security groups and references to resources created by the stack template.
If you specify a network interface, you must specify any security groups as part of the network interface instead.

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-launchtemplate-launchtemplatedata.html

internal/deployers/eksapi/node.go
mselim00: good catch, we actually already do some pre-validation to make sure a single instance type is set so we wouldn't even make it this far. this can just be taken out
https://github.com/aws/aws-k8s-tester/blob/main/internal/deployers/eksapi/deployer.go#L289

internal/deployers/eksapi/templates/unmanaged-nodegroup.yaml.template
mselim00: I was just trying to keep as much as possible in native cloudformation but yeah this makes sense, will change that

internal/deployers/eksapi/node.go
```diff
@@ -471,21 +374,25 @@ func (m *nodeManager) createUnmanagedNodegroupWithEFA(infra *Infrastructure, clu
...
 		volumeMountPath = "/dev/xvdb"
 	}
-	instanceType := opts.InstanceTypes[0]
-	networkInterfaces, err := m.getEfaNetworkInterfaces(instanceType, []string{cluster.securityGroupId}, subnetId)
-	if err != nil {
-		return fmt.Errorf("error getting efa interfaces: %v", err)
+	var networkInterfaces []networkInterface
+	instanceTypes := opts.InstanceTypes
+	if opts.EFA {
+		networkInterfaces, err = m.getEfaNetworkInterfaces(instanceTypes[0], []string{cluster.securityGroupId}, subnetId)
```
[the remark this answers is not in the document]
mselim00: could perhaps be more explicit about it, but atm we do always print the instance types that the test will use
https://github.com/aws/aws-k8s-tester/blob/main/internal/deployers/eksapi/node.go#L162

and in the EFA case this will always be just that one instance type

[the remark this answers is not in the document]
ndbaker1: alright nice just wanted to be sure 👍

internal/deployers/eksapi/templates/unmanaged-nodegroup.yaml.template
ndbaker1: yea i would have agreed if we completely gotten rid of it 😢

internal/deployers/eksapi/templates/unmanaged-nodegroup.yaml.template
mselim00: switched to create a network interface in both cases so it's a bit clearer

internal/deployers/eksapi/templates/unmanaged-nodegroup.yaml.template
mattcjo: Yep, this is nice.

internal/deployers/eksapi/node.go
```diff
@@ -713,41 +617,85 @@ func (m *nodeManager) getSubnetWithCapacity(infra *Infrastructure, opts *deploye
...
-			},
-			{
-				Name:   aws.String("subnet-id"),
-				Values: infra.subnetsPrivate,
-			},
+	klog.Infof("Using capacity reservation: %s", aws.ToString(capacityReservation.CapacityReservationId))
+	return capacityReservation, nil
+}
+
+func (m *nodeManager) getValidAvailabilityZonesFilter(opts *deployerOptions, infra *Infrastructure) ([]string, error) {
```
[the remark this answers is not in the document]
mattcjo: This feature would currently require quite a bit of refactoring in order to support it. I'm good with saving that enhancement for a later time.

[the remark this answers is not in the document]
mselim00: Yeah this would be nice to have. Another alternative would be creating an EC2 placement group and leaving EC2 to provision the group in whatever AZ it can and then creating the network interfaces based on that

Both these get complicated though. In this case though, we're only discovering the AZ in the create nodegroup function. Ideally we'd want to implement the retry at the level above this, but that'd complicate things as we currently don't discover a lot of information that would be needed to make that call until we're in this function.

I think `ec2:DescribeInstanceTypeOfferings` isn't a guaranteed ordering with respect to AZs though, so this should probably also work given enough retries of the entire deployer.

mselim00:
Tested 4 unmanaged nodegroup creation cases to confirm nodes create and join the cluster

1. Non-efa non-capacity reservation
```
kubetest2 eksapi --kubernetes-version 1.30 --region us-east-1 --instance-types m4.large m5.xlarge m6i.large m6i.xlarge --unmanaged-nodes --node-creation-timeout 10m --node-ready-timeout 10m --ami ami-0c7b87309c6d6af12 --nodes=2 --up --emit-metrics --user-data-format nodeadm --addons=vpc-cni:latest --test=exec --test=multi -- --fail-fast=false -- ginkgo ginkgo --use-binaries-from-path --parallel=6 --ginkgo-args=--no-color --focus-regex=\\[Conformance\\] --skip-regex='\\[Serial\\]|\\[Disruptive\\]|\\[Slow\\]|Garbage.collector' -- ginkgo --use-binaries-from-path --parallel=1 --ginkgo-args=--flake-attempts=3 --no-color --focus-regex='(\\[Serial\\]|\\[Disruptive\\]|\\[Slow\\]|Garbage.collector).*\\[Conformance\\]' --skip-regex=Daemon.set.+should.rollback.without.unnecessary.restarts
```
In this case, the ASG used all 4 subnets in the cluster. 

2. Non-efa capacity reservation
```
kubetest2 eksapi --kubernetes-version 1.30 --region us-east-1 --instance-types p4de.24xlarge --unmanaged-nodes --capacity-reservation --node-creation-timeout 10m --node-ready-timeout 10m --ami ami-07c1f432b61bc37f4 --nodes=1 --up --emit-metrics --user-data-format nodeadm --addons=vpc-cni:latest --test=exec -- nvidia.test --test.timeout=60m --test.v --test.run=^TestSingleNodeUnitTest$/unit-test --nvidiaTestImage=<IMAGE_URI> --nodeType=p4de.24xlarge
```
In this case, the ASG only used the 2 subnets in the AZ of the capacity reservation

3. Non-capacity reservation EFA
```
kubetest2 eksapi --kubernetes-version 1.30 --region us-east-1 --instance-types g6.8xlarge --unmanaged-nodes --node-creation-timeout 10m --node-ready-timeout 10m --ami ami-07c1f432b61bc37f4 --nodes=2 --efa --up --emit-metrics --user-data-format nodeadm --addons=vpc-cni:latest --test=exec -- nvidia.test --test.timeout=60m --test.v --test.run=^TestSingleNodeUnitTest$/unit-test --nvidiaTestImage=<IMAGE_URI> --nodeType=g6.8xlarge
```
In this case, the ASG used only the one private subnet from the AZ that was discovered by the`DescribeInstanceTypeOfferings` call

4. Capacity reservation EFA
```
kubetest2 eksapi --kubernetes-version 1.30 --region us-east-1 --instance-types trn1n.32xlarge --unmanaged-nodes --node-creation-timeout 10m --node-ready-timeout 10m --ami=ami-014d1c76173c85954 --nodes=4 --capacity-reservation --up --emit-metrics --user-data-format=nodeadm --addons=vpc-cni:latest --efa --test=exec -- neuron.test --test.timeout=60m --test.v --efaEnabled=true --nodeType=trn1n.32xlarge --neuronTestImage=<IMAGE_URI>
```
In this case, the ASG used only the one private subnet from the AZ of the capcaity reservation

mattcjo (approved):
LGTM

PR #590, opened 2025-02-27
Build private & public subnet in every AZ

mselim00:
*Issue #, if available:*

*Description of changes:*
Updates the infrastructure stack to include a private and public subnet in every availability zone in the region. Various capacity constraints make selecting two random availability zones, like we were doing before, too restrictive. For example, if a instance type is only offered in 2 out of 5 availability zones in a region, then those are also currently our odds of even including a subnet in the necessary AZ to try to launch the capacity. 

There's enough different scenarios to consider that we shouldn't try to pre-process like we were doing before with capacity reservations, to try to ensure that the particular AZ we need is included. That strategy also opens up weird failure modes. Since the infrastructure stack is created before cluster creation (which can take several minutes), a terminating instance in an ASG would not be counted for the subnet selection, but then we might try to target it if it is available by the time we start creating nodegroups. 

Unfortunately cloudformation currently has no native support for loops so I made this a go template, which we're already using for all other stacks at this point. 

At this point, I've only tested directly creating the stack to confirm that all resources are valid and unique where needed. I'll do some runs of a couple e2e scenarios before merging.


internal/deployers/eksapi/templates/templates.go
```diff
@@ -5,13 +5,21 @@ import (
...
 
-//go:embed infra.yaml
-var Infrastructure string
-
 var (
 	//go:embed unmanaged-nodegroup.yaml.template
 	unmanagedNodegroupTemplate string
 	UnmanagedNodegroup         = template.Must(template.New("unmanagedNodegroup").Parse(unmanagedNodegroupTemplate))
+
+	// for addition in a template
```
mattcjo, line 13: nit: This comment could be more descriptive about the components being added together.

internal/deployers/eksapi/templates/infra.yaml.template
```diff
@@ -114,47 +68,29 @@ Resources:
...
-        - Key: Name
-          Value:
-            Fn::Sub: "${AWS::StackName}/NATGateway01"
-  NATGateway02:
-    Type: AWS::EC2::NatGateway
-    DependsOn:
-      - NatGatewayEIP2
-      - SubnetPublic02
-      - VPCGatewayAttachment
+{{ range $index, $cidr := .PublicCIDRs }}
```
mattcjo, line 71: I think a comment here describing the subnet creation logic would be helpful. For example, something like 

```
# Create public subnets, one per AZ
# Each subnet gets a CIDR block from the calculated public CIDR ranges
```

mselim00:
I'll add the comments in a revision soon. Just one thing I'm concerned about with this change is all the resources that'll be created - particularly NAT Gateways. Some consumers may start exceeding service quotas if they run too many testes in parallel

mselim00:
Closing in favor of https://github.com/aws/aws-k8s-tester/pull/597

PR #594, opened 2025-03-13
fix(nvidia,containerd): skip ecr sandbox image check if localhost ima…

ndbaker1:
*Issue #, if available:*

*Description of changes:*

EKS AL2023 images will fail this check before the sandbox is built into the AMI as `localhost/kubernetes/pause`. add a branch to skip this check if the image name begins with `localhost`.

### DIFF
--- a/test/cases/nvidia/manifests/daemonset-containerd-check.yaml
+++ b/test/cases/nvidia/manifests/daemonset-containerd-check.yaml
@@ -40,11 +40,15 @@ spec:
             exit 1
           fi
 
-          # 4. Check that $sandbox_image references .ecr.
-          if [[ "$sandbox_image" != *".ecr."* ]]; then
-            echo "FAIL: no .ecr. reference in $sandbox_image"
-            echo "=== debug ==="
-            exit 1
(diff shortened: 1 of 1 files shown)


test/cases/nvidia/manifests/daemonset-containerd-check.yaml
```diff
@@ -40,8 +40,8 @@ spec:
             exit 1
           fi
 
-          # 4. Check that $sandbox_image references .ecr.
-          if [[ "$sandbox_image" != *".ecr."* ]]; then
+          # 4. Check that $sandbox_image references .ecr. or is provided on the instance
+          if [[ "$sandbox_image" != *".ecr."* ]] && [[ "$sandbox_image" != "localhost"* ]]; then
             echo "FAIL: no .ecr. reference in $sandbox_image"
```
mselim00, line 45: this failure message is a bit misleading given the check. maybe split off that case and log an INFO/WARN that the check is skipped?

ndbaker1: yea that's fair, ill split this out

mselim00 (approved):

PR #599, opened 2025-03-28
fix: best effort AZ selection for infra stack

mselim00:
*Issue #, if available:*

*Description of changes:*
We recently inadvertently added validation that all request instance types can be ran in the created VPC - which can be too strict when trying to run scarce instance types. 

All the requested instance types are added to an ASG which scales based on availability of the instance types - therefore we shouldn't directly validate that all the instance types can be ran in all the AZs, just that we're selecting the AZs where most instance types can be ran so that the ASG has the best change at scaling. This is particularly important for GPU instance types, where it might be preferable to pass a long list of instance types which may have zero overlap.

This also addresses the edge case of the requested instance type(s) only being offered in a single AZ, we should not fail in that case either.

### DIFF
--- a/internal/deployers/eksapi/infra.go
+++ b/internal/deployers/eksapi/infra.go
@@ -42,7 +42,7 @@ const (
 )
 
 // this value is not currently configurable, the infra stack is hardcoded to create 2
-const numAZs = 2
+const numInfraAZs = 2
 
 // eksEndpointURLTag is the key for an optional tag on the infrastructure CloudFormation stack,
 // which indicates which EKS environment is associated with the stack's resources.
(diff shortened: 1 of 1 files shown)


wwvela (approved):
chat offline. LGMT. We probably could handle the instance type picking which is supported in region as well here in future.

internal/deployers/eksapi/infra.go
```diff
@@ -406,11 +414,12 @@ func (m *InfrastructureManager) getAZsWithInstanceTypes(opts *deployerOptions) (
...
-	for az, count := range counts {
-		if count == len(opts.InstanceTypes) {
-			azs = append(azs, az)
-		}
+	for az := range counts {
+		azs = append(azs, az)
 	}
+	sort.Slice(azs, func(i, j int) bool {
+		return counts[azs[i]] > counts[azs[j]]
+	})
```
cartermckinnon, line 422: I think this makes sense, but I think we should also explicitly filter out AZs that do not support any of the specified instance types, it'll just make the logic more reliable if/when we change `numAZs`.

internal/deployers/eksapi/infra.go
```diff
@@ -99,28 +104,30 @@ func (m *InfrastructureManager) createInfrastructureStack(opts *deployerOptions)
...
 		}
 	}
+	// make sure we always have 2 AZs b/c the infra stack always tries to create 2
+	// can end up here if using a single capacity reservation or instance type(s) only offered in one AZ
+	for _, az := range azs.AvailabilityZones {
+		if len(subnetAzs) == numAZs {
+			break
+		}
+		if !slices.Contains(subnetAzs, *az.ZoneName) {
+			subnetAzs = append(subnetAzs, *az.ZoneName)
```
cartermckinnon, line 128: Can you add a log line here?

wwvela (approved):

PR #600, opened 2025-04-08
Add --no-asg eksapi deployer option

mselim00:
*Issue #, if available:*

*Description of changes:*

Launching and tearing down an ASG can take longer than just managing the individual instances, so this change adds an option to just create the bare instances to support use cases where that could make a difference.


cartermckinnon:
I'm not sure this is going to be meaningfully faster, and ASG solves all sorts of problems (ICE, AZ balancing, mixed instance types, etc.).

> to support use cases where that could make a difference.

Do you have something specific in mind?

PR #608, opened 2025-04-23
build(e2e): update image versions and add cuda compute 10.0

ndbaker1:
*Issue #, if available:*

*Description of changes:*

This PR makes the following changes:
* placed all frequently reused manifests into a `manifests` package to avoid having to store/embed files in multiple test cases
* update version of the neuron and nvidia device plugins
* bump the version of pytorch to 2.6 and build with compute capability 10.0


mselim00:
thanks for the refactoring/cleanup and 18k duplicate LOC saved!

CI's are failing to build the nvidia images though. nvidia-infernece just seems to be a space issue on the runner, are we possibly able to drop some of the older cuda archs from the supported list? I can't tell why the training one is failing though, that just seems unexpected.

mselim00 (approved):

PR #610, opened 2025-05-01
fix(nvidia/hpc-benchmarks): run job with 1 completion

mselim00:
*Issue #, if available:*

*Description of changes:*

We run this job with mpirun creating 1 process for each GPU. I really don't see why we would want 20 completions here, and functionally that takes a bit over an hour to complete - which means runs of this are likely to just timeout since most users would apply a 60 minute or less timeout. 


ndbaker1 (approved):
there's no completion count based how the [image overview runs it directly](https://catalog.ngc.nvidia.com/orgs/nvidia/containers/hpc-benchmarks), so lgtm

PR #611, opened 2025-05-01
feat(al2023): enable DRA in kubelet for 1.33

ndbaker1:
*Issue #, if available:*

*Description of changes:*

add kubelet featureGate to `NodeConfig` if the kubernetes version is 1.33 which applies only to AL2023 nodes created by the `eksapi` deployer.

### DIFF
--- a/internal/deployers/eksapi/templates/templates.go
+++ b/internal/deployers/eksapi/templates/templates.go
@@ -66,6 +66,7 @@ type UserDataTemplateData struct {
 	CertificateAuthority string
 	CIDR                 string
 	APIServerEndpoint    string
+	KubeletFeatureGates  map[string]bool
 }
 
 var (
--- a/internal/deployers/eksapi/templates/userdata_nodeadm.yaml.mimepart.template
+++ b/internal/deployers/eksapi/templates/userdata_nodeadm.yaml.mimepart.template
@@ -10,3 +10,11 @@ spec:
     apiServerEndpoint: {{.APIServerEndpoint}}
     certificateAuthority: {{.CertificateAuthority}}
     cidr: {{.CIDR}}
+{{- if .KubeletFeatureGates}}
+  kubelet:
+    config:
+      featureGates:
+        {{- range $gate, $value := .KubeletFeatureGates }}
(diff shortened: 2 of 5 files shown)


mselim00 (approved):

dims:
@ndbaker1 @mselim00 Karpenter doesn't yet understand DRA https://github.com/kubernetes-sigs/karpenter/issues/1231 - should we keep it optional for now?

internal/deployers/eksapi/templates/userdata_nodeadm.yaml.mimepart.template
```diff
@@ -10,3 +10,9 @@ spec:
     apiServerEndpoint: {{.APIServerEndpoint}}
     certificateAuthority: {{.CertificateAuthority}}
     cidr: {{.CIDR}}
+{{- if eq .KubernetesVersion "1.33" }}
```
mselim00, line 13:
```suggestion
{{- if ge .KubernetesVersion "1.33" }}
```

internal/deployers/eksapi/templates/templates.go
```diff
@@ -66,6 +66,7 @@ type UserDataTemplateData struct {
 	CertificateAuthority string
 	CIDR                 string
 	APIServerEndpoint    string
+	KubernetesVersion    string
```
cartermckinnon, line 69: could you pass in `KubeletFeatureGates map[string]bool` instead? Avoiding the conditional in the template would be nice IMO

ndbaker1: true 👍

internal/deployers/eksapi/templates/userdata_nodeadm.yaml.mimepart.template
mselim00: closing in favor of feature gate suggestion below

internal/deployers/eksapi/templates/templates.go
ndbaker1: this should confine the scope better per @dims point

PR #556, opened 2025-01-09
Fix sigs.k8s.io/kubetest2 version ref

cartermckinnon:
*Description of changes:*

`d7fcb799ce84ceda66c8b9b1ec8eefcbe226f293` -> `v0.0.0-20231113220322-d7fcb799ce84`

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #614, opened 2025-05-09
Update the README

wwvela:
*Description of changes:*
* update the README with latest binary name


README.md
```diff
@@ -4,6 +4,7 @@
 
 This project will use rolling releases going forward; we recommend fetching the latest commit:
 ```
+go install sigs.k8s.io/kubetest2/...@v0.0.0-20231113220322-d7fcb799ce84
```
wwvela, line 7: This is to install binary in kubetest2 upstream like kubetest2-noop, kubetest2-kind

README.md
```diff
@@ -4,6 +4,7 @@
...
 go install github.com/aws/aws-k8s-tester/...@HEAD
```
wwvela, line 8: This will only install the kubetest2 binary defined in our repo

README.md
```diff
@@ -26,8 +27,7 @@ This deployer is a thin wrapper around `eksctl`.
 
 The simplest usage is:
 ```
-kubetest2 \
-  eksctl \
+kubetest2-eksctl \
```
ndbaker1, line 30: these were correct before, kubetest2 picks the deployer and tester names from you're environment `PATH` up using the `kubetest2-...` format

README.md
ndbaker1: i think there was a conflict in the past due to some dependency, but do we still need this now?
https://github.com/aws/aws-k8s-tester/blob/2dd8edef17fed9612b82439920be9d428618323a/Dockerfile#L19

README.md
wwvela: Yeah It is correct only when we do `go install sigs.k8s.io/kubetest2` as well and it will install kubetest2 binary in go env. If we only run install of this pakcage `go install github.com/aws/aws-k8s-tester/...@HEAD` we will only have what we defined in cmd folder. So at the beginning I though the command should changed. 

I think we only need to add `go install sigs.k8s.io/kubetest2` in the top install command and it should be good

README.md
wwvela: I found we do this in our makefile to avoid the flack so I reuse it here. https://github.com/aws/aws-k8s-tester/pull/556 But it might only needed for the image build I guess. If we just build in local `go install sigs.k8s.io/kubetest2` should be enough

README.md
```diff
@@ -4,7 +4,7 @@
 
 This project will use rolling releases going forward; we recommend fetching the latest commit:
 ```
-go install github.com/aws/aws-k8s-tester/...@HEAD
+go install sigs.k8s.io/kubetest2 && go install github.com/aws/aws-k8s-tester/...@HEAD
 ```
```
cartermckinnon, line 8:
````suggestion
go install github.com/aws/aws-k8s-tester/...@HEAD
```

You'll need the standard `kubetest` tools as well:
```
go install sigs.k8s.io/kubetest2/...@latest
```
````

cartermckinnon (approved):

PR #617, opened 2025-05-19
feat: add option to skip node readiness

mselim00:
*Issue #, if available:*

*Description of changes:*

This adds an option to avoid blocking on nodes becoming ready in the deployer, this can be helpful for troubleshooting node joining issues, or for testing in Auto mode if you do not need to block on capacity being immediately available.


tzneal (approved):
lgtm

PR #620, opened 2025-05-21
calculate neuron capacity after NeuronCoreCapacity discovered

wwvela:
*Issue #, if available:*
* calculate neuron capacity after NeuronCoreCapacity discovered

### DIFF
--- a/test/cases/neuron-inference/main_test.go
+++ b/test/cases/neuron-inference/main_test.go
@@ -64,6 +64,7 @@ func TestMain(m *testing.M) {
 			return ctx, nil
 		},
 		discoverNeuronCoreCapacity,
+		getNodeCapacity
 	)
 
 	// Finish steps: remove device plugin if desired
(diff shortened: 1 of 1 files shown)


internal/deployers/eksapi/infra.go
```diff
@@ -30,7 +30,7 @@ import (
 const (
 	infraStackCreationTimeout         = time.Minute * 15
 	infraStackDeletionTimeout         = time.Minute * 30
-	networkInterfaceDetachmentTimeout = time.Minute * 5
+	networkInterfaceDetachmentTimeout = time.Minute * 10
```
shvbsle, line 33: why do we increase this?

wwvela: I saw some neuron training test pass but the ECS task failed because of the leaked ENI clean up timeout. So bump up the wait time to avoid flacky test

test/cases/neuron-inference/main_test.go
```diff
@@ -115,3 +117,41 @@ func discoverNeuronCoreCapacity(ctx context.Context, config *envconf.Config) (co
...
+			totalNeuronCore += int(neuronCoreCap.Value())
+		} else {
+			log.Printf("[WARN] Node %s (type=%s) lacks 'aws.amazon.com/neuroncore'.", node.Name, instanceType)
+		}
+	}
+	nodeCount := len(nodes.Items)
+	if nodeCount > 0 {
+		neuronPerNode = totalNeuron / nodeCount
+		neuronCorePerNode = totalNeuronCore / nodeCount
+	}
```
mselim00, line 154: we should probably just return an error here if nodeCount in the else condition

test/cases/neuron-inference/main_test.go
```diff
@@ -115,3 +117,41 @@ func discoverNeuronCoreCapacity(ctx context.Context, config *envconf.Config) (co
...
+		} else {
+			log.Printf("[WARN] Node %s (type=%s) lacks 'aws.amazon.com/neuron'.", node.Name, instanceType)
+		}
+		if hasNeuronCore {
+			totalNeuronCore += int(neuronCoreCap.Value())
+		} else {
+			log.Printf("[WARN] Node %s (type=%s) lacks 'aws.amazon.com/neuroncore'.", node.Name, instanceType)
+		}
+	}
+	nodeCount := len(nodes.Items)
```
mselim00, line 150: can you scope this math to nodes with `instanceType` == `nodeType` (including the nodeCount)?

We only schedule pods for this test to nodes of that type, so if there are other nodes in the cluster with neuron capacity, considering them could cause other issues.

https://github.com/aws/aws-k8s-tester/blob/main/test/cases/neuron-training/manifests/bert-training.yaml#L42

wwvela: Yeah make sense. Revised

test/cases/neuron-inference/main_test.go
wwvela: Revised

mselim00 (approved):

PR #626, opened 2025-06-03
feat: remove node affinities for EFA device plugin

mselim00:
*Issue #, if available:*

*Description of changes:*
Removing the hardcoded instance types to lighten the maintenance load as new instance types are constantly being released. Currently, the primary container in the plugin should just fast exit if there are no devices on the node with a message like
```
2025/06/03 23:18:52 Fetching EFA devices.
2025/06/03 23:18:52 No devices found. The plugin is either incompatible with this node or has failed to identify an EFA device. Restart pod after an EFA device has been attached, or prevent plugin from being scheduled on this node using a node selector.
```


ndbaker1 (approved):
i agree the overhead isn't so useful when its just a test resource

mselim00:
merging this - looks like the change was uncontroversial. this change is super isolated so it'll be easy enough to revert if we encounter a test scenario where this is problematic

PR #627, opened 2025-06-04
feat: add quick test to verify node topology labels

mselim00:
*Issue #, if available:*

*Description of changes:*
Adds a simple test case to verify node topology labels against the expected values from the ec2 DescribeInstanceTopology API

Sample output from testing in a cluster with 2 `t3.medium` and a single `p3dn.24xlarge`
```
2025/06/04 03:14:52 Starting quick test suite...
=== RUN   TestNodeTopology
=== RUN   TestNodeTopology/node-topology
=== RUN   TestNodeTopology/node-topology/Use_default_resources_limit
    node_topology_test.go:54: checking instance topologies for 1 nodes (out of 3 nodes in the cluster)
    node_topology_test.go:60: verifying instance topology for node ip-172-31-29-184.us-west-2.compute.internal (type: p3dn.24xlarge)
--- PASS: TestNodeTopology (0.93s)
    --- PASS: TestNodeTopology/node-topology (0.93s)
        --- PASS: TestNodeTopology/node-topology/Use_default_resources_limit (0.93s)
PASS
ok      github.com/aws/aws-k8s-tester/test/cases/quick  0.953s
```


test/cases/quick/node_topology_test.go
```diff
@@ -0,0 +1,87 @@
...
+	topology := features.New("node-topology").
+		WithLabel("suite", "node-topology").
+		Assess("Nodes have correct network topology labels", func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
+
+			clientset, err := kubernetes.NewForConfig(cfg.Client().RESTConfig())
+			if err != nil {
+				t.Fatalf("failed to create Kubernetes client: %v", err)
+			}
+
+			nodes, err := clientset.CoreV1().Nodes().List(ctx, metav1.ListOptions{})
```
ndbaker1: should be able to just use the `envconf.Config` to get a Kubernetes client for simple ops like List nodes

test/cases/quick/node_topology_test.go
```diff
@@ -0,0 +1,87 @@
...
+				instanceType := node.Labels["node.kubernetes.io/instance-type"]
+
+				t.Logf("verifying instance topology for node %s (type: %s)", node.Name, instanceType)
+
+				for i, networkNode := range nodeTopology.NetworkNodes {
+					// https://github.com/kubernetes/cloud-provider-aws/blob/b47d2cf2a33ae655cd353ec42ea43362b804c397/pkg/providers/v1/well_known_labels.go#L26
+					expectedLabel := cloudprovider.LabelNetworkNodePrefix + strconv.Itoa(i+1)
+					if actualValue, ok := node.Labels[expectedLabel]; !ok {
+						t.Logf("node %s (type: %s) does not have expected network label %s", node.Name, instanceType, expectedLabel)
+						t.Fail()
```
ndbaker1:
```suggestion
						t.Errorf("node %s (type: %s) does not have expected network label %s", node.Name, instanceType, expectedLabel)
```

test/cases/quick/node_topology_test.go
```diff
@@ -0,0 +1,76 @@
...
+
+	"github.com/aws/aws-k8s-tester/internal/e2e"
+	"github.com/aws/aws-sdk-go-v2/aws"
+	v1 "k8s.io/api/core/v1"
+	cloudprovider "k8s.io/cloud-provider-aws/pkg/providers/v1"
+	"sigs.k8s.io/e2e-framework/pkg/envconf"
+	"sigs.k8s.io/e2e-framework/pkg/features"
+)
+
+func TestNodeTopology(t *testing.T) {
```
cartermckinnon, line 20: Since the topology labels only exist for a handful of instance types (mostly nvidia ones) maybe we should just move this to the nvidia suite? Or do you plan to filter it out when the `quick` suite is used for unsupported instance types?

mselim00: The test should be safe to run on any cluster (assuming the tester role has the correct permissions). The DescribeInstanceTopology API only returns responses for the subset of the requested instances that have data, and then only the corresponding nodes are validated in this test. So in clusters without any supported nodes, the test would just fast succeed after getting an empty response from the API. 

It should be a relatively fast and cheap check, so I think it’s valuable in the quick suite to get the non-nvidia coverage

cartermckinnon: okay cool, I didn't remember if this API returned an error if the instanceId you gave it was for an unsupported instance type

ndbaker1 (approved):

PR #631, opened 2025-06-24
Filter out local zones when selecting AZs in infra.go

annapoornanarayan:
For `Error: failed to wait for infrastructure stack creation: waiter state transitioned to Failure` on creating cluster due to IPv6 CIDR not being supported in local AZs.
*Issue #, if available:*

*Description of changes:*. 
Added filtering in `DescribeAvailabilityZones` to only request regular availability zones (`zone-type=availability-zone`). Modified `counts` to consider regular AZs only.

### DIFF
--- a/internal/deployers/eksapi/infra.go
+++ b/internal/deployers/eksapi/infra.go
@@ -95,17 +95,17 @@ func (m *InfrastructureManager) createInfrastructureStack(opts *deployerOptions)
 	// TODO: create a subnet in every AZ
 	// get two AZs for the subnets
 	azs, err := m.clients.EC2().DescribeAvailabilityZones(context.TODO(), &ec2.DescribeAvailabilityZonesInput{
-    Filters: []ec2types.Filter{
-        	{
+		Filters: []ec2types.Filter{
+			{
 				Name:   aws.String("zone-type"),
 				Values: []string{opts.ZoneType},
-        	},
-    	},
+			},
+		},
 	})
 	if err != nil {
 		return nil, err
 	}
-	
+
 	var allowedAZs []string
 	for i := 0; i < numInfraAZs; i++ {
 		allowedAZs = append(allowedAZs, aws.ToString(azs.AvailabilityZones[i].ZoneName))
@@ -428,7 +428,7 @@ func (m *InfrastructureManager) getRankedAZsForInstanceTypes(opts *deployerOptio
 			{
 				Name:   aws.String("location"),
 				Values: allowedAZs,
-        	},
+			},
 		},
 	})
 	if err != nil {
--- a/internal/deployers/eksapi/infra.go
+++ b/internal/deployers/eksapi/infra.go
@@ -91,10 +91,57 @@ func (i *Infrastructure) subnets() []string {
 	return append(i.subnetsPublic, i.subnetsPrivate...)
 }
 
+// getAZsWithIPv6Support returns availability zones that support IPv6 CIDR blocks
+func (m *InfrastructureManager) getAZsWithIPv6Support() (*ec2.DescribeAvailabilityZonesOutput, error) {
+	// Get all AZs that are either regular AZs or IPv6-compatible Local Zones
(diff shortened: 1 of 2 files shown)


wwvela:
* We might need to add filter for [DescribeAvailabilityZones here as well](https://github.com/aws/aws-k8s-tester/blob/main/internal/deployers/eksapi/infra.go#L97C2-L100), for case if there is no instance type passed.
* The IPv6 CIDR is generated by default here in [infra template](https://github.com/aws/aws-k8s-tester/blob/main/internal/deployers/eksapi/templates/infra.yaml#L87-L92), so we must pick AZs which support the IPv6 CIDR. Unless we add more flags to make it an optional. To quickly unblock the test now, filter local zone out is acceptable

internal/deployers/eksapi/infra.go
```diff
@@ -417,12 +417,34 @@ func (m *InfrastructureManager) getRankedAZsForInstanceTypes(opts *deployerOptio
...
+		Filters: []ec2types.Filter{
+			{
+				Name:   aws.String("zone-type"),
+				Values: []string{"availability-zone"},
+			},
+		},
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to describe availability zones: %v", err)
+	}
```
wwvela: * I feel this is kind of duplicate here https://github.com/aws/aws-k8s-tester/blob/main/internal/deployers/eksapi/infra.go#L97C2-L100. You can add the filter in above api call to filter out the local zone and pass it into `getRankedAZsForInstanceTypes` method. Is can also cover the case if no instanceTypes flag passed in command
* And the [DescribeInstanceTypeOfferingsInput](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/ec2#DescribeInstanceTypeOfferingsInput) allows you pass filter `location - The location. For a list of possible identifiers`, so you can pass the AZs in filter to get those AZs availble for instance type only

mselim00: just adding on to here, we could also perhaps add a new flag to control the "zone type" and make that default to availability-zone. that would allow us to still deploy in a local zone if we so chose, and I think it generally makes sense that we would want one zone type for the test resources.

 this also seems to be a requirement for allocating an IPv6 CIDR because local zones are in a different network border group than the rest of the region, e.g. for us-west-2 AZs are part of the "us-west-2" group, but local zones in lax-1 are in "us-west-2-lax-1", which leads to the error `The IPv6 CIDR <some-cidr> is not valid for the provided Availability Zone. You must use an IPv6 CIDR from the us-west-2-lax-1 network border group.` when trying to use a CIDR from the us-west-2 group in us-west-2-lax-1

annapoornanarayan:
Summary of changes:
1. deployer.go:
- Added flag `ZoneType` in `deployerOptions `struct which defaults to `availability-zone` (`VerifyUpFlags()` ).
2. infra.go: 
- Filtered for `ZoneType` in `createInfrastructureStack` and added valid AZs to new list allowedAZs
- If initialization takes [InstanceType branch](https://github.com/aws/aws-k8s-tester/blob/386d5790bb62fb126066d1286fd0f176602fec7a/internal/deployers/eksapi/infra.go#L107) , it will now take allowedAZs as an argument into `getRankedAZsForInstanceTypes`.
- Within `getRankedAZsForInstanceTypes`, use allowedAZs to only get instances within valid AZs using filter on `location`.
- Also, revert `counts` to original structure

internal/deployers/eksapi/infra.go
```diff
@@ -114,7 +127,7 @@ func (m *InfrastructureManager) createInfrastructureStack(opts *deployerOptions)
 		}
 		subnetAzs = azs[0:int(math.Min(float64(len(azs)), numInfraAZs))]
 	} else {
-		for i := range numInfraAZs {
+		for i := 0; i < numInfraAZs; i++ {
 			subnetAzs = append(subnetAzs, aws.ToString(azs.AvailabilityZones[i].ZoneName))
```
mselim00, line 131: nit: this could just be allowedAZs[i] now

wwvela (approved):

Issue #595, opened 2025-03-19
eksctl-deployer: has limited input args and does not support config file


ytsssun:
The `aws-k8s-tester` eksctl deployer is a wrapper around eksctl, but it currently exposes only basic functionality and lacks support for many useful eksctl features. During my testing with `eksapi` deployer, I found it less stable than manually using `eksctl` for deployment. Enriching the current wrapper of eksctl as the deployer seems a more feasible way to ensure a more stabled deploying -> testing workflow.

### Current Limitations
The current eksctl deployer wrapper has significant limitations:
- No support for using custom eksctl config files directly
- Limited cluster configuration options (forced cluster naming, limited nodegroup options)
- Missing support for advanced networking like EFA and node configuration
- Unable to leverage many of eksctl's [powerful features](https://eksctl.io/usage/schema/)

### Proposed Enhancement
Enhance the `eksctl` deployer to support more configuration options and allow direct use of eksctl config files, making it a more complete wrapper around eksctl's capabilities.

This will improve flexibility when testing EKS clusters and remove the need for users to switch to different tools when more advanced eksctl features are needed.

PR #636, opened 2025-07-01
feat(deployer/eksctl): allow deleting nodegroupds with --down, skip draining nodegroups option

maherthomsi:
*Issue #, if available:*
[#595 ](https://github.com/aws/aws-k8s-tester/issues/595)
*Description of changes:*

This PR adds the capability to **delete** EKS nodegroups using the --down flag. The changes include:

- Added a new SkipDrain flag option to control node draining behavior during deletion
  - The nodegroup deletion supports an optional --drain=false flag to skip draining nodes before deletion
- Enhanced the Down() function to support two deployment targets:
  - nodegroup: Deletes a specific nodegroup from a cluster
  - cluster: Deletes the entire EKS cluster (existing behavior)
- Added proper error handling and logging for both nodegroup and cluster deletion operations

Tests Done
```
kubetest2 eksctl \
--cluster-name=my-cluster-130 \
--nodegroup-name=aarch64-aws-k8s-130 \
--region=us-west-2 \
--kubernetes-version=1.30 \
--ami-family=Bottlerocket \
--ami=ami-017b2a28378e53647 \
--instance-types=m7g.xlarge \
--nodes=3 \
--deploy-target nodegroup \
--skip-drain \
--up \
--down
```
```
kubetest2 eksctl \
--cluster-name=my-cluster-130 \
--nodegroup-name=aarch64-aws-k8s-130 \
--region=us-west-2 \
--kubernetes-version=1.30 \
--ami-family=Bottlerocket \
--ami=ami-017b2a28378e53647 \
--instance-types=m7g.xlarge \
--nodes=3 \
--deploy-target nodegroup \
--skip-drain \
--up \
--down \
--test ginkgo -- \
--focus-regex='\[Conformance\]' \
--parallel=6 \
--skip-regex='\[Serial\]|\[Disruptive\]|\[Slow\]|Garbage.collector'
```

### DIFF
--- a/internal/deployers/eksctl/down.go
+++ b/internal/deployers/eksctl/down.go
@@ -14,8 +14,7 @@ func (d *deployer) Down() error {
 
 	if d.DeployTarget == "nodegroup" {
 		klog.Infof("deleting nodegroup %s from cluster %s", d.NodegroupName, d.clusterName)
-		args := []string{"delete", "nodegroup", "--cluster", d.clusterName, "--name", d.NodegroupName, "--drain=false", "--wait"}
-		err = util.ExecuteCommand("eksctl", args...)
+		err = util.ExecuteCommand("eksctl", "delete", "nodegroup", "--cluster", d.clusterName, "--name", d.NodegroupName, "--drain=false", "--wait")
 		if err != nil {
 			return fmt.Errorf("failed to delete nodegroup: %v", err)
 		}
--- a/internal/deployers/eksctl/down.go
+++ b/internal/deployers/eksctl/down.go
@@ -1,16 +1,38 @@
 package eksctl
 
 import (
+	"fmt"
+
 	"github.com/aws/aws-k8s-tester/internal/util"
 	"k8s.io/klog"
 )
 
 func (d *deployer) Down() error {
-	klog.Infof("deleting cluster %s", d.commonOptions.RunID())
-	err := util.ExecuteCommand("eksctl", "delete", "cluster", "--name", d.commonOptions.RunID(), "--wait")
-	if err != nil {
-		return err
+	d.initClusterName()
+
+	var err error
+
+	if d.DeployTarget == "nodegroup" {
+		klog.Infof("deleting nodegroup %s from cluster %s", d.NodegroupName, d.clusterName)
+		args := []string{"delete", "nodegroup", "--cluster", d.clusterName, "--name", d.NodegroupName}
+		if d.SkipDrain {
+			args = append(args, "--drain=false")
+		}
+		args = append(args, "--wait")
+		err = util.ExecuteCommand("eksctl", args...)
+		if err != nil {
+			return fmt.Errorf("failed to delete nodegroup: %v", err)
+		}
+		klog.Infof("Successfully deleted nodegroup: %s from cluster: %s", d.NodegroupName, d.clusterName)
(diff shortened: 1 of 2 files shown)


internal/deployers/eksctl/down.go
```diff
@@ -1,16 +1,38 @@
...
+	d.initClusterName()
+
+	var err error
+
+	if d.DeployTarget == "nodegroup" {
+		klog.Infof("deleting nodegroup %s from cluster %s", d.NodegroupName, d.clusterName)
+		args := []string{"delete", "nodegroup", "--cluster", d.clusterName, "--name", d.NodegroupName}
+		if d.SkipDrain {
+			args = append(args, "--drain=false")
+		}
```
mselim00: still trying to understand this aspect, do you think you could provide an example use case where this would be desirable? as a wrapper it would generally best to keep the exposed flag profile from the source binary minimal, otherwise the routing logic can quickly get out of control

internal/deployers/eksctl/down.go
```diff
@@ -1,16 +1,38 @@
...
+		args = append(args, "--wait")
```
mselim00: nit: this could just be in the initial array

internal/deployers/eksctl/down.go
maherthomsi: During conformance testing, test namespaces get marked for deletion while pods are still running. When eksctl tries to drain nodes, it can't create eviction objects in terminating namespaces, causing the deletion to fail with errors like:
```
2025-06-26 14:25:19 [!]  pod eviction error ("unrecoverable error evicting pod: job-8955/indexed-job-0-9bllr: pods \"indexed-job-0-9bllr\" is forbidden: unable to create new content in namespace job-8955 because it is being terminated") on node ip-192-168-83-225.us-west-2.compute.internal
2025-06-26 14:25:24 [!]  error occurred during drain, to skip drain use '--drain=false' flag
Error: draining nodegroups: unrecoverable error evicting pod: job-8955/indexed-job-0-9bllr: pods "indexed-job-0-9bllr" is forbidden: unable to create new content in namespace job-8955 because it is being terminated
Error: exit status 255
```
We would need to skip drianing nodes for our testing to be able to delete these nodegroups and we aren't concerned about evicting pods.  Furthermore, when terminating nodegroups we don't care if they are gracefully closed and skipping the draining process would save us time.

I do think your concern with complicated routing logic is valid. We can simplify by defaulting to skip draining when deleting nodegroups for testing. This streamlines the process and aligns with our testing needs, while reducing routing complexity. We can still retain this option to drain if explicitly required by another use case.

internal/deployers/eksctl/down.go
```diff
@@ -1,16 +1,34 @@
...
-	err := util.ExecuteCommand("eksctl", "delete", "cluster", "--name", d.commonOptions.RunID(), "--wait")
-	if err != nil {
-		return err
+	d.initClusterName()
+
+	var err error
+
+	if d.DeployTarget == "nodegroup" {
+		klog.Infof("deleting nodegroup %s from cluster %s", d.NodegroupName, d.clusterName)
+		args := []string{"delete", "nodegroup", "--cluster", d.clusterName, "--name", d.NodegroupName, "--drain=false", "--wait"}
```
mselim00: nit: this can just be in directly in the execute command call now

mselim00 (approved):

PR #639, opened 2025-07-02
feat(nvidia): run multiple multi-node perfs

ndbaker1:
*Issue #, if available:*

*Description of changes:*

This PR increases coverage of different perf nccl-tests. adding all 3 of `all_reduce`, `all_gather`, and `alltoall` in the same multi-node test.

It also ensures that logs are printed from the job even when the job has not completed by a new, upfront deadline of 10 minutes (tunable in the future) and prints the job logs even on failure.

### DIFF
--- a/test/cases/nvidia/mpi_test.go
+++ b/test/cases/nvidia/mpi_test.go
@@ -7,7 +7,9 @@ import (
 	_ "embed"
 	"fmt"
 	"regexp"
+	"strings"
 	"testing"
+	"time"
 
 	fwext "github.com/aws/aws-k8s-tester/internal/e2e"
 	"github.com/aws/aws-k8s-tester/internal/e2e/mpijobs"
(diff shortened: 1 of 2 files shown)


test/cases/nvidia/mpi_test.go
```diff
@@ -139,26 +98,30 @@ func TestMPIJobPytorchTraining(t *testing.T) {
 			return ctx
 		}).
 		Assess("MPIJob succeeds", func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
-			mpiJob := mpijobs.NewUnstructured("multi-node-nccl-test", "default")
+			mpiJob := mpijobs.NewUnstructured(jobName, "default")
 			t.Log("Waiting for multi node job to complete")
+			ctxJobSucceed, _ := context.WithTimeout(ctx, 10*time.Minute)
```
mselim00: this might be low for the initial run where we have to wait on the image pull and then potential delays in the launcher starting after the workers w/ the crash loop backoff, maybe we could double it to 20?

ndbaker1: i see the concern but 20 minutes for each is still a lot. you're worried about the upfront pull delays and ~couple minutes of synchronization but the tests generally run fast in my experience as of late.

the misleading bit is that the first test will cache the image, and then subsequent tests might only take 5-10 minutes realistically. if they spin during that time then there would be an issue with the test rather than the image or setup

ndbaker1: feel free to change it if that's your only concern, we can always tune it afterwards

mselim00 (approved):

PR #648, opened 2025-07-09
CloudWatch infra support for eksapi deployer

annapoornanarayan:
*Issue #, if available:*

*Description of changes:*

Adds CloudWatch monitoring infrastructure setup to the eksapideployer, enabling automated deployment of CloudWatch agents and required IAM resources.

Files Modified:
- templates/templates.go - Added CloudWatch template definitions and data structures
- templates/cloudwatch-infra.yaml.template - OIDC provider and IAM role template
- templates/cloudwatch_agent_infra.yaml - Agent namespace, ClusterRole, and bindings
- cluster.go - Added OIDC issuer URL extraction
- infra.go - CloudWatch stack creation/deletion functions
- deployer.go - Integrated CloudWatch stacks after cluster and nodegroup formation and before deletion

Testing:
 - Verified CloudWatch infrastructure creates successfully
 - Confirmed proper cleanup on cluster deletion in Cloud Formation


internal/deployers/eksapi/infra.go
```diff
@@ -475,3 +475,98 @@ func (m *InfrastructureManager) getAZsWithCapacity(opts *deployerOptions) ([]str
...
+func (m *InfrastructureManager) createCloudWatchInfrastructureStack(clusterName, oidcIssuerURL string) (string, error) {
+	clusterUUID := strings.TrimPrefix(clusterName, "kubetest2-eksapi-")
+	stackName := fmt.Sprintf("%s-cloudwatch", m.resourceID)
+	
+	oidcProviderURL := strings.TrimPrefix(oidcIssuerURL, "https://")
+	templateData := templates.CloudWatchInfraTemplateData{
+		ClusterName:     clusterName,
+		OIDCIssuerURL:   oidcIssuerURL,
+		OIDCProviderURL: oidcProviderURL,
+		ResourceId:      clusterUUID,
```
wwvela: Instead of ResourceId, use clusterUUID directly can be more easily for other to understand what is this param from

internal/deployers/eksapi/deployer.go
```diff
@@ -231,6 +233,27 @@ func (d *deployer) Up() error {
 			// don't return err, this isn't critical
 		}
 	}
+	
+	if d.deployerOptions.StaticClusterName == "" {
```
wwvela: The `cloudwatch agent infra` is just deploy service role, and cluster role binding and access required by cloud watch agent. It doesn't seems related to if using static cluster or not. You can deploy the cloudwatch agent infra for all cluster by default.

internal/deployers/eksapi/deployer.go
```diff
@@ -231,6 +233,27 @@ func (d *deployer) Up() error {
...
+		klog.Infof("Setting up CloudWatch infrastructure...")
+		oidcIssuerURL := d.cluster.oidcIssuerURL
+		if roleArn, err := d.infraManager.createCloudWatchInfrastructureStack(d.cluster.name, oidcIssuerURL); err != nil {
+			klog.Errorf("CloudWatch infrastructure setup failed: %v", err)
+			return err
+		} else {
+			d.cluster.cloudwatchRoleArn = roleArn
+			klog.Infof("CloudWatch infrastructure setup completed after node creation")
+		}
```
wwvela: Doesn't seems the else required here. We can do `d.cluster.cloudwatchRoleArn, err := d.infraManager.createCloudWatchInfrastructureStack(d.cluster.name, oidcIssuerURL)` directly? And the log `CloudWatch infrastructure setup completed after node creation` is not needed, as the CW infra creation doesn't rely on node creation process.

internal/deployers/eksapi/deployer.go
```diff
@@ -231,6 +233,27 @@ func (d *deployer) Up() error {
...
+		// Apply CloudWatch infrastructure manifest
+		if err := d.applyCloudWatchInfraManifest(); err != nil {
+			klog.Errorf("CloudWatch infrastructure manifest failed: %v", err)
+			return err
+		} else {
+			klog.Infof("CloudWatch infrastructure manifest applied successfully")
+		}
```
wwvela: can be move out from else

internal/deployers/eksapi/infra.go
```diff
@@ -475,3 +475,98 @@ func (m *InfrastructureManager) getAZsWithCapacity(opts *deployerOptions) ([]str
 	}
 	return subnetAzs, nil
 }
+
+func (m *InfrastructureManager) createCloudWatchInfrastructureStack(clusterName, oidcIssuerURL string) (string, error) {
+	clusterUUID := strings.TrimPrefix(clusterName, "kubetest2-eksapi-")
+	stackName := fmt.Sprintf("%s-cloudwatch", m.resourceID)
```
wwvela: The stack name should be `cloudwatch-<clusteruuid>`, It will be more easily to catch the purpose of the stack by name immediately. the uuid is quite long

internal/deployers/eksapi/infra.go
```diff
@@ -475,3 +475,98 @@ func (m *InfrastructureManager) getAZsWithCapacity(opts *deployerOptions) ([]str
...
+	var templateBuffer strings.Builder
+	if err := templates.CloudWatchInfra.Execute(&templateBuffer, templateData); err != nil {
+		return "", fmt.Errorf("failed to execute CloudWatch infrastructure template: %w", err)
+	}
+	
+	renderedTemplate := templateBuffer.String()
+	
+	input := cloudformation.CreateStackInput{
+		StackName:    aws.String(stackName),
+		TemplateBody: aws.String(renderedTemplate),
```
wwvela: nit: this can be `aws.String(templateBuffer.String())`

internal/deployers/eksapi/infra.go
```diff
@@ -475,3 +475,98 @@ func (m *InfrastructureManager) getAZsWithCapacity(opts *deployerOptions) ([]str
...
+			klog.Infof("CloudWatch infrastructure stack created successfully with role ARN: %s", *output.OutputValue)
+			return *output.OutputValue, nil
+		}
+	}
+	
+	return "", fmt.Errorf("CloudWatch role ARN not found in stack outputs")
+}
+
+func (m *InfrastructureManager) deleteCloudWatchInfrastructureStack() error {
+	stackName := fmt.Sprintf("%s-cloudwatch", m.resourceID)
```
wwvela: nit: same stack name issue

internal/deployers/eksapi/infra.go
```diff
@@ -475,3 +475,98 @@ func (m *InfrastructureManager) getAZsWithCapacity(opts *deployerOptions) ([]str
...
+				StackName: aws.String(stackName),
+			},
+			infraStackDeletionTimeout)
+	if err != nil {
+		klog.Warningf("failed to wait for CloudWatch infrastructure stack deletion: %v", err)
+		return nil
+	}
+	
+	klog.Infof("deleted CloudWatch infrastructure stack: %s", stackName)
+	return nil
```
wwvela: I feel there is no need to wait the CW infra stack fually clean up as this will further extend the test runtime

internal/deployers/eksapi/templates/cloudwatch-infra.yaml.template
```diff
@@ -0,0 +1,50 @@
...
+Resources:
+  OIDCProvider:
+    Type: AWS::IAM::OIDCProvider
+    Properties:
+      Url: {{ .OIDCIssuerURL }}
+      ClientIdList:
+        - sts.amazonaws.com
+      Tags:
+        - Key: Name
+          Value: {{ .ResourceId }}-oidc-provider
```
wwvela: You can use the Pod Agent Identity instead of OIDCProvider which is a kind of old way to implement it. It can also be created using cloud formation template kind of similar as the OIDC but you don't need to get the OIDCIssuerURL, but use the cluser name instead. something like below: 

```
TestPodIdentityAssociation:
    Type: AWS::EKS::PodIdentityAssociation
    Properties:
      ClusterName: !Ref ClusterName
      RoleArn: !Ref RoleArn
      Namespace: my-namespace
      ServiceAccount: !Ref ServiceAccount`
```

internal/deployers/eksapi/infra.go
```diff
@@ -475,3 +475,86 @@ func (m *InfrastructureManager) getAZsWithCapacity(opts *deployerOptions) ([]str
 	}
 	return subnetAzs, nil
 }
+
+func (m *InfrastructureManager) createCloudWatchInfrastructureStack(clusterName, oidcIssuerURL string) (string, error) {
+	clusterUUID := strings.TrimPrefix(clusterName, "kubetest2-eksapi-")
+	stackName := fmt.Sprintf("cloudwatch-%s", clusterUUID)
+	
+	oidcProviderURL := strings.TrimPrefix(oidcIssuerURL, "https://")
```
wwvela: There is no need to do these string parse if using pod identity agent instead of OIDC Provider

internal/deployers/eksapi/templates/cloudwatch_agent_infra.yaml
```diff
@@ -0,0 +1,52 @@
...
+metadata:
+  name: amazon-cloudwatch
+  labels:
+    name: amazon-cloudwatch
+
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: cwagent-prometheus
```
wwvela: I think we should avoid `-prometheus` in the property name which is kind of missleading as we didn't really deploy prometheus, but only deployed the cloudwatch agent which can read any metric source with prometheus format

internal/deployers/eksapi/deployer.go
```diff
@@ -73,8 +75,8 @@ type deployerOptions struct {
 	IPFamily                string        `flag:"ip-family" desc:"IP family for the cluster (ipv4 or ipv6)"`
 	KubeconfigPath          string        `flag:"kubeconfig" desc:"Path to kubeconfig"`
 	KubernetesVersion       string        `flag:"kubernetes-version" desc:"cluster Kubernetes version"`
-	LogBucket               string        `flag:"log-bucket" desc:"S3 bucket for storing logs for each run. If empty, logs will not be stored."`
 	NodeadmFeatureGates     []string      `flag:"nodeadm-feature-gates" desc:"Feature gates to enable for nodeadm (key=value pairs)"`
+	LogBucket               string        `flag:"log-bucket" desc:"S3 bucket for storing logs for each run. If empty, logs will not be stored."`
```
wwvela: why this move in the middle of flags starts with 'N', Seems the flags was in alphabetical order

internal/deployers/eksapi/templates/templates.go
```diff
@@ -8,12 +8,21 @@ import (
 //go:embed infra.yaml
 var Infrastructure string
 
+//go:embed cloudwatch_agent_infra.yaml
+var CloudWatchAgentInfra string
```
wwvela: Could you change this to `CloudWatchAgentRbac` as it is all rbac related resources. And use `xxinfra` is quite similar as` cloudwatch-infra.yaml.template` below

internal/deployers/eksapi/infra.go
```diff
@@ -475,3 +475,81 @@ func (m *InfrastructureManager) getAZsWithCapacity(opts *deployerOptions) ([]str
...
 	return subnetAzs, nil
 }
+
+func (m *InfrastructureManager) createCloudWatchInfrastructureStack(clusterName string) (string, error) {
+	stackName := fmt.Sprintf("cloudwatch-%s", strings.TrimPrefix(clusterName, "kubetest2-eksapi-"))
+
+	var templateBuffer strings.Builder
+	err := templates.CloudWatchInfra.Execute(&templateBuffer, templates.CloudWatchInfraTemplateData{
+		ClusterName: clusterName,
+		ClusterUUID: stackName,
```
wwvela: Is it ClusterUUID or the cloudwatch stack name? Saw you pass the stack name directly. And this role name will have duplicate `cloudwatch` in name if passing stackName
```
RoleName: cloudwatch-role-{{ .ClusterUUID }}
```

internal/deployers/eksapi/infra.go
```diff
@@ -475,3 +475,81 @@ func (m *InfrastructureManager) getAZsWithCapacity(opts *deployerOptions) ([]str
...
+		}
+	}
+
+	return "", fmt.Errorf("CloudWatch role ARN not found in stack outputs")
+}
+
+func (m *InfrastructureManager) deleteCloudWatchInfrastructureStack() error {
+	// We need to reconstruct the clusterUUID from resourceID to match the creation pattern
+	clusterUUID := strings.TrimPrefix(m.resourceID, "kubetest2-eksapi-")
+	stackName := fmt.Sprintf("cloudwatch-%s", clusterUUID)
```
wwvela: Can pass strings.TrimPrefix(m.resourceID, "kubetest2-eksapi-") to stackName directly? clusterUUID is not needed anywhere else in this method

wwvela:
Leave some nit comments, Overall LGTM

internal/deployers/eksapi/infra.go
annapoornanarayan: You're right! It will get duplicated. Fixing this

internal/deployers/eksapi/infra.go
```diff
@@ -475,3 +475,81 @@ func (m *InfrastructureManager) getAZsWithCapacity(opts *deployerOptions) ([]str
...
 	return subnetAzs, nil
 }
+
+func (m *InfrastructureManager) createCloudWatchInfrastructureStack(clusterName string) (string, error) {
+	stackName := fmt.Sprintf("cloudwatch-%s", strings.TrimPrefix(clusterName, "kubetest2-eksapi-"))
+
+	var templateBuffer strings.Builder
+	err := templates.CloudWatchInfra.Execute(&templateBuffer, templates.CloudWatchInfraTemplateData{
+		ClusterName: clusterName,
+		ClusterUUID: strings.TrimPrefix(clusterName, "kubetest2-eksapi-"),
```
wwvela: clusterUUID := strings.TrimPrefix(m.resourceID, "kubetest2-eksapi-")
as the uuid is used in multi place

internal/deployers/eksapi/infra.go
```diff
@@ -475,3 +475,81 @@ func (m *InfrastructureManager) getAZsWithCapacity(opts *deployerOptions) ([]str
...
+		}
+	}
+
+	return "", fmt.Errorf("CloudWatch role ARN not found in stack outputs")
+}
+
+func (m *InfrastructureManager) deleteCloudWatchInfrastructureStack() error {
+	// We need to reconstruct the clusterUUID from resourceID to match the creation pattern
+	clusterUUID := strings.TrimPrefix(m.resourceID, "kubetest2-eksapi-")
+	stackName := fmt.Sprintf("cloudwatch-%s", clusterUUID)
```
wwvela: uuid is only used in stack name, no need to assign to a param

wwvela (approved):

internal/deployers/eksapi/templates/cloudwatch_agent_infra.yaml
```diff
@@ -0,0 +1,50 @@
+apiVersion: v1
```
mselim00, line 1: nit: `templates/` has been CFN templates up until this point. can you make a separate directory e.g `internal/deployer/eksapi/manifests` for k8s resource templates like this one?

internal/deployers/eksapi/templates/cloudwatch-infra.yaml.template
```diff
@@ -0,0 +1,48 @@
+AWSTemplateFormatVersion: '2010-09-09'
+Description: kubetest2-eksapi CloudWatch using Pod Identity
+
+Resources:
+  CloudWatchRole:
+    Type: AWS::IAM::Role
+    Properties:
+      RoleName: cloudwatch-role-{{ .ClusterUUID }}
```
mselim00: simple string substitutions like this can just be done natively in cloudformation using parameters. we should try to keep as cfn-native, as possible and only use go templating for more advanced things that cfn doesn't support (e.g. looping) because it reduces portability

as an example, see `infra.yaml`. We create a parameter for the `ResourceID`
https://github.com/aws/aws-k8s-tester/blob/591301bac4188846767aa51bf34359844310e0e4/internal/deployers/eksapi/templates/unmanaged-nodegroup.yaml.template#L5-L8

and pass that in from the node manager
https://github.com/aws/aws-k8s-tester/blob/591301bac4188846767aa51bf34359844310e0e4/internal/deployers/eksapi/node.go#L392-L400

then use that to name the launch template (and others)
https://github.com/aws/aws-k8s-tester/blob/591301bac4188846767aa51bf34359844310e0e4/internal/deployers/eksapi/templates/unmanaged-nodegroup.yaml.template#L109

internal/deployers/eksapi/infra.go
```diff
@@ -475,3 +475,80 @@ func (m *InfrastructureManager) getAZsWithCapacity(opts *deployerOptions) ([]str
...
+		if errors.As(err, &notFound) {
+			klog.Infof("CloudWatch infrastructure stack does not exist: %s", stackName)
+			return nil
+		}
+		klog.Warningf("failed to delete CloudWatch infrastructure stack: %v", err)
+		return nil
+	}
+
+	klog.Infof("initiated deletion of CloudWatch infrastructure stack: %s", stackName)
+	return nil
```
mselim00: these should return errors instead of logging

internal/deployers/eksapi/deployer.go
```diff
@@ -338,6 +356,9 @@ func (d *deployer) Down() error {
 }
 
 func deleteResources(im *InfrastructureManager, cm *ClusterManager, nm *nodeManager, k8sClient *k8sClient /* nillable */, opts *deployerOptions /* nillable */) error {
+	if err := im.deleteCloudWatchInfrastructureStack(); err != nil {
+		klog.Warningf("Failed to delete CloudWatch infrastructure: %v", err)
+	}
```
mselim00: any reason why we're not propagating the error here? seems prone to allowing resource leaks if this just logs warnings without propagating a failure code

internal/deployers/eksapi/infra.go
```diff
@@ -475,3 +475,80 @@ func (m *InfrastructureManager) getAZsWithCapacity(opts *deployerOptions) ([]str
...
+	stackName := fmt.Sprintf("cloudwatch-%s", clusterUUID)
+
+	var templateBuffer strings.Builder
+	err := templates.CloudWatchInfra.Execute(&templateBuffer, templates.CloudWatchInfraTemplateData{
+		ClusterName: clusterName,
+		ClusterUUID: clusterUUID,
+	})
+	if err != nil {
+		return "", fmt.Errorf("executing CloudWatch template: %w", err)
+	}
```
mselim00:
```suggestion
	if err := templates.CloudWatchInfra.Execute(&templateBuffer, templates.CloudWatchInfraTemplateData{
		ClusterName: clusterName,
		ClusterUUID: clusterUUID,
	}); err != nil {
		return "", fmt.Errorf("executing CloudWatch template: %w", err)
	}
```
when the only return from a function is an error and it's not needed anywhere else, it's more idiomatic to wrap it's scope inside an `if` block that handles the error. there's a couple other places this could also be applied

internal/deployers/eksapi/infra.go
```diff
@@ -475,3 +475,80 @@ func (m *InfrastructureManager) getAZsWithCapacity(opts *deployerOptions) ([]str
...
+
+	return "", fmt.Errorf("CloudWatch role ARN not found in stack outputs")
+}
+
+func (m *InfrastructureManager) deleteCloudWatchInfrastructureStack() error {
+	stackName := fmt.Sprintf("cloudwatch-%s", strings.TrimPrefix(m.resourceID, "kubetest2-eksapi-"))
+
+	input := cloudformation.DeleteStackInput{
+		StackName: aws.String(stackName),
+	}
```
mselim00: nit: this is small enough that it could just be in the `DeleteStack` call

internal/deployers/eksapi/infra.go
```diff
@@ -475,3 +475,80 @@ func (m *InfrastructureManager) getAZsWithCapacity(opts *deployerOptions) ([]str
...
+	})
+	if err != nil {
+		return "", fmt.Errorf("failed to describe CloudWatch infrastructure stack: %w", err)
+	}
+
+	for _, output := range stack.Stacks[0].Outputs {
+		if *output.OutputKey == "CloudWatchRoleArn" {
+			klog.Infof("CloudWatch infrastructure stack created successfully with role ARN: %s", *output.OutputValue)
+			return *output.OutputValue, nil
+		}
```
mselim00:
```suggestion
		if aws.ToString(output.OutputKey) == "CloudWatchRoleArn" {
		        outputValue := aws.ToString(output.OutputValue)
			klog.Infof("CloudWatch infrastructure stack created successfully with role ARN: %s", outputValue)
			return outputValue, nil
		}
```
using functions like `aws.ToString` or `ptr.ToString` give free handling of potential `nil` pointers by returning an empty string instead of dereferencing if the pointer is nil

internal/deployers/eksapi/infra.go
```diff
@@ -475,3 +475,80 @@ func (m *InfrastructureManager) getAZsWithCapacity(opts *deployerOptions) ([]str
...
+	}
+
+	input := cloudformation.CreateStackInput{
+		StackName:    aws.String(stackName),
+		TemplateBody: aws.String(templateBuffer.String()),
+		Capabilities: []cloudformationtypes.Capability{cloudformationtypes.CapabilityCapabilityNamedIam},
+	}
+
+	klog.Infof("creating CloudWatch infrastructure stack: %s", stackName)
+	out, err := m.clients.CFN().CreateStack(context.TODO(), &input)
```
mselim00:
```suggestion
	out, err := m.clients.CFN().CreateStack(context.TODO(), cloudformation.CreateStackInput{
		StackName:    aws.String(stackName),
		TemplateBody: aws.String(templateBuffer.String()),
		Capabilities: []cloudformationtypes.Capability{cloudformationtypes.CapabilityCapabilityNamedIam},
	})
```
it's easier to trace if the input is in the function call

internal/deployers/eksapi/templates/templates.go
```diff
@@ -8,12 +8,21 @@ import (
 //go:embed infra.yaml
 var Infrastructure string
 
+//go:embed cloudwatch_agent_infra.yaml
+var CloudWatchAgentRbac string
```
mselim00:
```suggestion
var CloudWatchAgentRbac []byte
```
you can embed as a `string`, `[]byte`, or `embed.FS`, and in this case it seems like the variable is only used as `[]byte`
https://pkg.go.dev/embed

internal/deployers/eksapi/cluster.go
```diff
@@ -32,6 +32,7 @@ type Cluster struct {
 	arn                      string
 	name                     string
 	cidr                     string
+	cloudwatchRoleArn        string
```
mselim00: should this be in `Infrastructure` instead?
https://github.com/aws/aws-k8s-tester/blob/749a1dde778e16db4139d7efa15e8f182b890ddf/internal/deployers/eksapi/infra.go#L80-L88

mselim00 (commented):
thanks for putting this together!

internal/deployers/eksapi/infra.go
```diff
@@ -475,3 +475,80 @@ func (m *InfrastructureManager) getAZsWithCapacity(opts *deployerOptions) ([]str
...
+			klog.Infof("CloudWatch infrastructure stack created successfully with role ARN: %s", *output.OutputValue)
+			return *output.OutputValue, nil
+		}
+	}
+
+	return "", fmt.Errorf("CloudWatch role ARN not found in stack outputs")
+}
+
+func (m *InfrastructureManager) deleteCloudWatchInfrastructureStack() error {
+	stackName := fmt.Sprintf("cloudwatch-%s", strings.TrimPrefix(m.resourceID, "kubetest2-eksapi-"))
```
mselim00:
```suggestion
	stackName := fmt.Sprintf("%s-cloudwatch", m.resourceID)
```
to be more consistent with the unmanaged nodegroup stack name 
https://github.com/aws/aws-k8s-tester/blob/591301bac4188846767aa51bf34359844310e0e4/internal/deployers/eksapi/node.go#L555-L557

also worth creating a single function to generate this name to make sure the name create and delete use never diverge

internal/deployers/eksapi/infra.go
```diff
@@ -475,3 +475,80 @@ func (m *InfrastructureManager) getAZsWithCapacity(opts *deployerOptions) ([]str
 	}
 	return subnetAzs, nil
 }
+
+func (m *InfrastructureManager) createCloudWatchInfrastructureStack(clusterName string) (string, error) {
+	clusterUUID := strings.TrimPrefix(clusterName, "kubetest2-eksapi-")
+	stackName := fmt.Sprintf("cloudwatch-%s", clusterUUID)
```
mselim00:
```suggestion
	stackName := fmt.Sprintf("%s-cloudwatch", m.resourceID)
```

annapoornanarayan: I changed it according to this comment: https://github.com/aws/aws-k8s-tester/pull/648#discussion_r2196121447

mselim00: hmm I see. I think it'd be more consistent and searchable with this change but I don't feel too strongly about that part, but it should use `m.resourceID` as either prefix suffix

annapoornanarayan: m.resourceID is the entire cluster name which exceeds the char limit (Gives this error: Member must have length less than or equal to 64) for the role name, so I cannot add that as a prefix/suffix.

mselim00: oh interesting. in that case, can you re-use the below in the TrimPrefix instead of hardcoding it? https://github.com/aws/aws-k8s-tester/blob/01c363ff46f336116ca03dd9f449278f46bbcd3c/internal/deployers/eksapi/deployer.go#L26

annapoornanarayan: Oh yes! Will modify this

annapoornanarayan:
Thanks for the comments! Here is the summary of changes I made:
- Moved CW Role to Infrastructure struct instead of Cluster
- Created separate manifests folder
- Changed go substitution to CloudFormation string substitution
- Changed relevant warning logs to return errors in Cloudwatch infra creation and deletion
- Fixed recommended code style changes for code clarity

internal/deployers/eksapi/infra.go
```diff
@@ -475,3 +476,74 @@ func (m *InfrastructureManager) getAZsWithCapacity(opts *deployerOptions) ([]str
 	}
 	return subnetAzs, nil
 }
+
+func (m *InfrastructureManager) createCloudWatchInfrastructureStack(clusterName string) (string, error) {
+	clusterUUID := strings.TrimPrefix(clusterName, ResourcePrefix+"-")
+	stackName := fmt.Sprintf("cloudwatch-%s", clusterUUID)
```
mselim00: nit: this is done twice, so it should be in a singular fuction

internal/deployers/eksapi/templates/templates.go
```diff
@@ -5,15 +5,21 @@ import (
 	"text/template"
 )
 
-//go:embed infra.yaml
+//go:embed manifests/infra.yaml
 var Infrastructure string
 
+//go:embed manifests/cloudwatch_agent_infra.yaml
+var CloudWatchAgentRbac []byte
+
```
mselim00: I think this might be clearer as
```
eksapi/
  templates/
  manifests/
```
where `templates` is for cloudformation templates
and `manifests` is for k8s resource manifests
(so `infra.yaml` would also stay in `templates` since it's a cloudformation template)

annapoornanarayan: I will move infra.yaml back! So when I tried moving it outside of templates, it was unable to access the file using` //go:embed` with relative paths as it does not support `..`. ([refer here)](https://pkg.go.dev/embed#:%7E:text=The%20patterns%20are%20interpreted%20relative%20to%20the%20package%20directory%20containing%20the%20source%20file.%20The%20path%20separator%20is%20a%20forward%20slash%2C%20even%20on%20Windows%20systems.%20Patterns%20may%20not%20contain%20%E2%80%98.%E2%80%99%20or%20%E2%80%98..%E2%80%99%20or%20empty%20path%20elements%2C%20nor%20may%20they%20begin%20or%20end%20with%20a%20slash.:~:text=The%20//go%3Aembed%20directive%20accepts%20multiple%20space%2Dseparated%20patterns%20for%20brevity%2C%20but%20it%20can%20also%20be%20repeated%2C%20to%20avoid%20very%20long%20lines%20when%20there%20are%20many%20patterns.%20The%20patterns%20are%20interpreted%20relative%20to%20the%20package%20directory%20containing%20the%20source%20file.) If i were to add a file outside the scope of templates.go I would have to create functions to embed.FS the entire directory. Its definitely possible, but it started to look like a large chunk of unnecessary code.

annapoornanarayan: Actually you can ignore this comment, I think I found the issue!

annapoornanarayan: Discussed offline and decided to move all files within templates

PR #649, opened 2025-07-09
CloudWatch Agent integration for nvidia-training

annapoornanarayan:
*Issue #, if available:*

*Description of changes:*
NOTE: This will only work after CW Infra is also successfully deployed (https://github.com/aws/aws-k8s-tester/pull/648)
Changes include: 
- Added amiVariant, teamIdentifier, and KubernetesVersion flags with default versions enabled
- CloudWatch agent manifest configured to scrape Prometheus compatible DCGM GPU metrics with proper labeling
- Small change in DCGM to ensure headless service discovery
- Created additional function to scrape Region from node labels
- Applied manifest in main_test.go after rendering with required parameters once DCGM Exporter runs

Testing:

- Verified CloudWatch agent deploys successfully with pods showing that 90+ metrics are collected successfully
- Metrics are actively visible under ContainerInsights/Prometheus under the dimensions ["TestType", "KubernetesVersion", "Variant", "InstanceType", "TeamIdentifier"]

### DIFF
--- a/test/cases/nvidia-training/main_test.go
+++ b/test/cases/nvidia-training/main_test.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"os/signal"
 	"slices"
+	"strings"
 	"testing"
 	"time"
 
@@ -31,26 +32,60 @@ func TestMain(m *testing.M) {
 		log.Fatalf("failed to initialize test environment: %v", err)
 	}
 
+	// Validate and transform flags
+	*amiVariant = strings.ToLower(*amiVariant)
+	if *kubernetesVersion != "1.32" && *kubernetesVersion != "1.33" {
+		log.Fatalf("kubernetesVersion must be either 1.32 or 1.33, got: %s", *kubernetesVersion)
+	}
+	
+	manifestVersion := strings.Replace(*kubernetesVersion, ".", "", 1)
+
 	ctx, cancel := signal.NotifyContext(context.Background(), os.Interrupt)
 	defer cancel()
 	testenv = env.NewWithConfig(cfg).WithContext(ctx)
 
-	manifests := [][]byte{
+	region, err := getRegionFromNodes(ctx, cfg)
+	if err != nil {
+		log.Printf("Warning: failed to get region from nodes, using default us-west-2: %v", err)
+		region = "us-west-2"
+	}
+
+	// template data for CloudWatch Agent manifest
+	templateData := map[string]string{
+		"VERSION":       manifestVersion,
+		"VARIANT":       *amiVariant,
+		"INSTANCE_TYPE": *nodeType,
+		"REGION":        region,
+		"TEAM_IDENTIFIER": *teamIdentifier,
+		"TEST_NAME":     "nvidia.training",
+	}
+
+	renderedCloudWatchAgentManifest, err := fwext.RenderManifests(manifests.CloudWatchAgentManifest, templateData)
+	if err != nil {
+		log.Fatalf("failed to render CloudWatch Agent manifest: %v", err)
+	}
+
+	manifestsList := [][]byte{
 		manifests.NvidiaDevicePluginManifest,
 		manifests.MpiOperatorManifest,
 		manifests.EfaDevicePluginManifest,
 		manifests.DCGMExporterManifest,
+		renderedCloudWatchAgentManifest,
 	}
+
+	log.Printf("Using template variables: REGION=%s, VERSION=%s, VARIANT=%s, INSTANCE_TYPE=%s, TEAM_IDENTIFIER=%s, TEST_NAME=nvidia.training", 
+		region, manifestVersion, *amiVariant, *nodeType, *teamIdentifier)
 
 	testenv.Setup(
 		// Apply all manifests
 		func(ctx context.Context, config *envconf.Config) (context.Context, error) {
-			log.Println("Applying NVIDIA device plugin, MPI operator, EFA device plugin and DCGM Exporter manifests.")
-			err := fwext.ApplyManifests(config.Client().RESTConfig(), manifests...)
+			log.Println("Applying NVIDIA device plugin, MPI operator, EFA device plugin, DCGM Exporter and CloudWatch Agent manifests.")
+
+			err := fwext.ApplyManifests(config.Client().RESTConfig(), manifestsList...)
 			if err != nil {
 				return ctx, fmt.Errorf("failed to apply manifests: %w", err)
 			}
-			log.Println("Successfully applied NVIDIA device plugin, MPI operator, EFA device plugin and DCGM Exporter manifests.")
+			log.Println("Successfully applied NVIDIA device plugin, MPI operator, EFA device plugin, DCGM Exporter and CloudWatch Agent manifests.")
 			return ctx, nil
 		},
 
@@ -77,19 +112,19 @@ func TestMain(m *testing.M) {
 		deployDaemonSet("nvidia-device-plugin-daemonset", "kube-system"),
 		deployDaemonSet("aws-efa-k8s-device-plugin-daemonset", "kube-system"),
 		deployDaemonSet("dcgm-exporter", "kube-system"),
-
+		deployDaemonSet("cwagent-prometheus", "amazon-cloudwatch"),
 		checkNodeTypes, // Dynamically check node types and capacities after device plugins are ready
 	)
 
 	testenv.Finish(
 		func(ctx context.Context, config *envconf.Config) (context.Context, error) {
-			log.Println("Deleting NVIDIA device plugin, MPI operator, EFA device plugin and DCGM Exporter manifests.")
-			slices.Reverse(manifests)
-			err := fwext.DeleteManifests(config.Client().RESTConfig(), manifests...)
+			log.Println("Deleting NVIDIA device plugin, MPI operator, EFA device plugin DCGM Exporter and CloudWatch Agent manifests.")
+			slices.Reverse(manifestsList)
+			err := fwext.DeleteManifests(config.Client().RESTConfig(), manifestsList...)
 			if err != nil {
 				return ctx, fmt.Errorf("failed to delete manifests: %w", err)
 			}
-			log.Println("Successfully deleted NVIDIA device plugin, MPI operator, EFA device plugin and DCGM Exporter manifests.")
+			log.Println("Successfully deleted NVIDIA device plugin, MPI operator, EFA device plugin, DCGM Exporter and CloudWatch Agent manifests.")
 			return ctx, nil
 		},
 	)
@@ -155,6 +190,20 @@ func checkNodeTypes(ctx context.Context, config *envconf.Config) (context.Contex
 	log.Printf("[INFO] EFA Per Node: %d", efaPerNode)
 
 	return ctx, nil
+}
+
+// getRegionFromNodes extracts the AWS region from node labels
+func getRegionFromNodes(ctx context.Context, config *envconf.Config) (string, error) {
+	clientset, err := kubernetes.NewForConfig(config.Client().RESTConfig())
+	if err != nil {
+		return "", fmt.Errorf("failed to create Kubernetes client: %w", err)
+	}
+	nodes, err := clientset.CoreV1().Nodes().List(ctx, metav1.ListOptions{Limit: 1})
+	if err != nil {
+		return "", fmt.Errorf("failed to list nodes: %w", err)
+	}
+	az := nodes.Items[0].Labels["topology.kubernetes.io/zone"]
+	return strings.TrimSuffix(az, az[len(az)-1:]), nil
 }
 // Helper function to deploy DaemonSet + Wait for Ready
 func deployDaemonSet(name, namespace string) env.Func {
--- a/test/cases/nvidia-training/vars.go
+++ b/test/cases/nvidia-training/vars.go
@@ -14,6 +14,9 @@ var (
 	bertTrainingImage *string
 	efaEnabled        *bool
 	nodeType          *string
+	kubernetesVersion *string
+	amiVariant	  	  *string
+	teamIdentifier    *string
 
 	nodeCount  int
 	gpuPerNode int
@@ -24,4 +27,7 @@ func init() {
 	bertTrainingImage = flag.String("bertTrainingImage", "", "Docker image used for BERT training workload")
 	efaEnabled = flag.Bool("efaEnabled", false, "Enable Elastic Fabric Adapter (EFA)")
 	nodeType = flag.String("nodeType", "", "Instance type for cluster nodes")
+	kubernetesVersion = flag.String("kubernetesVersion", "1.32", "Kubernetes version for the cluster")
+	amiVariant = flag.String("amiVariant", "al2023", "AMI variant for the cluster nodes")
+	teamIdentifier = flag.String("teamIdentifier", "node-runtime", "Team identifier for resource tagging")
 }
--- a/test/manifests/assets/cloudwatch-agent.yaml
+++ b/test/manifests/assets/cloudwatch-agent.yaml
@@ -1,0 +1,130 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: prometheus-cwagentconfig
+  namespace: amazon-cloudwatch
+data:
+  cwagentconfig.json: |
+    {
+      "agent": {
+        "debug": true,
+        "region": "{{.REGION}}"
+      },
+      "logs": {
+        "metrics_collected": {
+          "prometheus": {
+            "prometheus_config_path": "/etc/prometheusconfig/prometheus.yaml",
+            "emf_processor": {
+              "metric_declaration": [
+                {
+                  "source_labels": ["job"],
+                  "label_matcher": "dcgm-exporter",
+                  "dimensions": [["TestType", "KubernetesVersion", "Variant", "InstanceType", "TeamIdentifier"]],
+                  "metric_selectors": [
+                    "^DCGM_FI_DEV_GPU_UTIL$",
+                    "^DCGM_FI_DEV_MEM_COPY_UTIL$",
+                    "^DCGM_FI_DEV_FB_USED$",
+                    "^DCGM_FI_DEV_FB_FREE$",
+                    "^DCGM_FI_DEV_POWER_USAGE$"
+                  ]
+                }
+              ]
+            }
+          }
+        },
+        "force_flush_interval": 5
+      }
+    }
+
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: prometheus-config
+  namespace: amazon-cloudwatch
+data:
+  prometheus.yaml: |
+    global:
+      scrape_interval: 1m
+      scrape_timeout: 10s
+    scrape_configs:
+      - job_name: dcgm-exporter
+        static_configs:
+          - targets:
+            - dcgm-exporter.kube-system.svc.cluster.local:9400
+        metrics_path: /metrics
+        scrape_interval: 30s
+        metric_relabel_configs:
+          - action: replace
+            source_labels: [job]
+            target_label: job
+            replacement: "dcgm-exporter"
+          - action: replace
+            source_labels: [job]
+            target_label: KubernetesVersion
+            replacement: "{{.VERSION}}"
+          - action: replace
+            source_labels: [job]
+            target_label: Variant
+            replacement: "{{.VARIANT}}"
+          - action: replace
+            source_labels: [job]
+            target_label: InstanceType
+            replacement: "{{.INSTANCE_TYPE}}"
+          - action: replace
+            source_labels: [job]
+            target_label: TeamIdentifier
+            replacement: "{{.TEAM_IDENTIFIER}}"
+          - action: replace
+            source_labels: [job]
+            target_label: TestType
+            replacement: "{{.TEST_NAME}}"
+
+
+
+
+---
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: cwagent-prometheus
+  namespace: amazon-cloudwatch
+spec:
+  selector:
+    matchLabels:
+      app: cwagent-prometheus
+  template:
+    metadata:
+      labels:
+        app: cwagent-prometheus
+    spec:
+      serviceAccountName: cwagent-prometheus
+      dnsPolicy: ClusterFirst 
+      containers:
+        - name: cloudwatch-agent
+          image: public.ecr.aws/cloudwatch-agent/cloudwatch-agent:latest
+          imagePullPolicy: Always
+          env:
+            - name: RUN_WITH_IRSA
+              value: "True"
+          resources:
+            limits:
+              cpu: 1000m
+              memory: 1000Mi
+            requests:
+              cpu: 200m
+              memory: 200Mi
+          volumeMounts:
+            - name: prometheus-cwagentconfig
+              mountPath: /etc/cwagentconfig
+            - name: prometheus-config
+              mountPath: /etc/prometheusconfig
+      volumes:
+        - name: prometheus-cwagentconfig
+          configMap:
+            name: prometheus-cwagentconfig
+        - name: prometheus-config
+          configMap:
+            name: prometheus-config
+      terminationGracePeriodSeconds: 60
+---
--- a/test/manifests/assets/dcgm-exporter.yaml
+++ b/test/manifests/assets/dcgm-exporter.yaml
@@ -57,6 +57,7 @@ metadata:
     app.kubernetes.io/name: "dcgm-exporter"
     app.kubernetes.io/version: "4.1.3"
 spec:
+  clusterIP: "None"
   selector:
     app.kubernetes.io/name: "dcgm-exporter"
     app.kubernetes.io/version: "4.1.3"
--- a/test/manifests/raw.go
+++ b/test/manifests/raw.go
@@ -21,4 +21,7 @@ var (
 	//go:embed assets/dcgm-exporter.yaml
 	DCGMExporterManifest []byte
 
+	//go:embed assets/cloudwatch-agent.yaml
+	CloudWatchAgentManifest []byte
+
 )


annapoornanarayan:
Changed scrape interval and improved CW Config for clarity

test/cases/nvidia-training/main_test.go
```diff
@@ -156,6 +191,20 @@ func checkNodeTypes(ctx context.Context, config *envconf.Config) (context.Contex
...
+func getRegionFromNodes(ctx context.Context, config *envconf.Config) (string, error) {
+	clientset, err := kubernetes.NewForConfig(config.Client().RESTConfig())
+	if err != nil {
+		return "", fmt.Errorf("failed to create Kubernetes client: %w", err)
+	}
+	nodes, err := clientset.CoreV1().Nodes().List(ctx, metav1.ListOptions{Limit: 1})
+	if err != nil {
+		return "", fmt.Errorf("failed to list nodes: %w", err)
+	}
+	az := nodes.Items[0].Labels["topology.kubernetes.io/zone"]
```
wwvela: Should use `topology.kubernetes.io/region` label directly instead of doing some parse based on zone label

test/cases/nvidia-training/main_test.go
```diff
@@ -31,26 +32,60 @@ func TestMain(m *testing.M) {
...
+
 	ctx, cancel := signal.NotifyContext(context.Background(), os.Interrupt)
 	defer cancel()
 	testenv = env.NewWithConfig(cfg).WithContext(ctx)
 
-	manifests := [][]byte{
+	region, err := getRegionFromNodes(ctx, cfg)
+	if err != nil {
+		log.Printf("Warning: failed to get region from nodes, using default us-west-2: %v", err)
+		region = "us-west-2"
```
wwvela: What will happen to CW agent If the test is not running in PDX but failed to get region info here and use PDX by defaut?

test/cases/nvidia-training/main_test.go
```diff
@@ -31,26 +32,60 @@ func TestMain(m *testing.M) {
 		log.Fatalf("failed to initialize test environment: %v", err)
 	}
 
+	// Validate and transform flags
+	*amiVariant = strings.ToLower(*amiVariant)
+	if *kubernetesVersion != "1.32" && *kubernetesVersion != "1.33" {
+		log.Fatalf("kubernetesVersion must be either 1.32 or 1.33, got: %s", *kubernetesVersion)
+	}
```
wwvela: version limit is not needed. We are able to run the test with other versions if we wanted and shouldn't avoid it. We only run 1.32 and 1.33 in our pipeline just because of limited GPU instance capacity resources.

test/cases/nvidia-training/main_test.go
```diff
@@ -31,26 +32,60 @@ func TestMain(m *testing.M) {
...
 		manifests.NvidiaDevicePluginManifest,
 		manifests.MpiOperatorManifest,
 		manifests.EfaDevicePluginManifest,
 		manifests.DCGMExporterManifest,
+		renderedCloudWatchAgentManifest,
 	}
 
+	log.Printf("Using template variables: REGION=%s, VERSION=%s, VARIANT=%s, INSTANCE_TYPE=%s, TEAM_IDENTIFIER=%s, TEST_NAME=nvidia.training", 
+		region, manifestVersion, *amiVariant, *nodeType, *teamIdentifier)
+
```
wwvela: nit: didn't see very helpful to print these out in log, we can saw these info in metric dimension directly

test/cases/nvidia-training/vars.go
```diff
@@ -24,4 +27,7 @@ func init() {
 	bertTrainingImage = flag.String("bertTrainingImage", "", "Docker image used for BERT training workload")
 	efaEnabled = flag.Bool("efaEnabled", false, "Enable Elastic Fabric Adapter (EFA)")
 	nodeType = flag.String("nodeType", "", "Instance type for cluster nodes")
+	kubernetesVersion = flag.String("kubernetesVersion", "1.32", "Kubernetes version for the cluster")
+	amiVariant = flag.String("amiVariant", "al2023", "AMI variant for the cluster nodes")
+	teamIdentifier = flag.String("teamIdentifier", "node-runtime", "Team identifier for resource tagging")
```
wwvela: The teamIdentifier default value should set to `unknown`, as someone who run this test in local and forget to pass this flag will emit their test metric datapoint in our team or any other team metric dashboards.

test/cases/nvidia-training/main_test.go
```diff
@@ -31,26 +32,60 @@ func TestMain(m *testing.M) {
...
+	region, err := getRegionFromNodes(ctx, cfg)
+	if err != nil {
+		log.Printf("Warning: failed to get region from nodes, using default us-west-2: %v", err)
+		region = "us-west-2"
+	}
+
+	// template data for CloudWatch Agent manifest
+	templateData := map[string]string{
+		"VERSION":       manifestVersion,
+		"VARIANT":       *amiVariant,
```
wwvela: can pass `strings.ToLower(*amiVariant)` directly here

test/cases/nvidia-training/main_test.go
```diff
@@ -31,26 +32,60 @@ func TestMain(m *testing.M) {
...
-	manifests := [][]byte{
+	region, err := getRegionFromNodes(ctx, cfg)
+	if err != nil {
+		log.Printf("Warning: failed to get region from nodes, using default us-west-2: %v", err)
+		region = "us-west-2"
+	}
+
+	// template data for CloudWatch Agent manifest
+	templateData := map[string]string{
+		"VERSION":       manifestVersion,
```
wwvela: can pass `strings.Replace(*kubernetesVersion, ".", "", 1)` directly

test/cases/nvidia-training/main_test.go
annapoornanarayan: If it fails to get the region the CW Agent won't write the metrics on CW. I was thinking since the Agent is not critical to the test itself to simply leave it as a warning, but I can make it a log.Fatalf so it prevents from proceeding without the correct region.

annapoornanarayan:
Changes: 
- Used topology.kubernetes.io/region label directly
- Failure on region info if not found
- Better k8 version validation
- Clean up logs
- "unknown" default value for teamIdentifier
- Passed in validation of flags directly to map

test/cases/nvidia-training/main_test.go
wwvela: Yeah. I think should give some context about it in log like where the metric will be emitted to. like `Warning: failed to get region from nodes, the test metrics will be emitted to PDX by default: %v`

test/cases/nvidia-training/main_test.go
```diff
@@ -156,6 +184,21 @@ func checkNodeTypes(ctx context.Context, config *envconf.Config) (context.Contex
...
+	clientset, err := kubernetes.NewForConfig(config.Client().RESTConfig())
+	if err != nil {
+		return "", fmt.Errorf("failed to create Kubernetes client: %w", err)
+	}
+	nodes, err := clientset.CoreV1().Nodes().List(ctx, metav1.ListOptions{Limit: 1})
+	if err != nil {
+		return "", fmt.Errorf("failed to list nodes: %w", err)
+	}
+	region := nodes.Items[0].Labels["topology.kubernetes.io/region"]
+	return region, nil
```
wwvela: nit: can return `nodes.Items[0].Labels["topology.kubernetes.io/region"]` here directly

test/cases/nvidia-training/vars.go
```diff
@@ -14,6 +14,9 @@ var (
 	bertTrainingImage *string
 	efaEnabled        *bool
 	nodeType          *string
+	kubernetesVersion *string
+	amiVariant	  *string
+	teamIdentifier    *string
```
wwvela: Was in a rabbit hole. Actually we can create node lables for these info easily in userdata in phase 1 resource generation, so all nodes will have these info in labels. And in phase 2 testing parts, we can just getting these info from node labels. 2 of below flags are duplicate as we can get them from deployer flags
* Only need to add one new `teamIdentifier` flag in deployer options 
* already have kubernetesVersion flag in deployer options
* amiVariant can be get according to the `--user-data-format`, if it is bootstrap, then the variant is al2, if it is al2023

We can generate the node lables in userdata:
* For nodeadm template example is like below
```
kubelet:
    config:
       ....
    flags:
    - --node-labels=app=my-app,environment=production
```

For bootstrap template example is like below:
```
/etc/eks/bootstrap.sh cluster_name \
--kubelet-extra-args '--node-labels=something=hello,somethingelse=bye'
```

annapoornanarayan: Spoke offline - making separate PR for this change

test/cases/nvidia-training/main_test.go
```diff
@@ -156,6 +185,20 @@ func checkNodeTypes(ctx context.Context, config *envconf.Config) (context.Contex
...
+	clientset, err := kubernetes.NewForConfig(config.Client().RESTConfig())
+	if err != nil {
+		return "", fmt.Errorf("failed to create Kubernetes client: %w", err)
+	}
+	nodes, err := clientset.CoreV1().Nodes().List(ctx, metav1.ListOptions{Limit: 1})
+	if err != nil {
+		return "", fmt.Errorf("failed to list nodes: %w", err)
+	}
+	return nodes.Items[0].Labels["topology.kubernetes.io/region"], nil
+}
```
mselim00: this is fine, but an alternative would have been using the AWS config
```
import "github.com/aws/aws-sdk-go-v2/config"
...
if cfg, err := config.LoadDefaultConfig(context.TODO()); err != nil {
   return "", fmt.Errorf("failed loading config, %v", err)
else {
   return cfg.Region, nil
}
```
though since we don't currently use any AWS APIs in this test, there's been no expectation that the environment is configured correctly. I think it could be worth adding a TODO to load it from the AWS default config if AWS API calls are added in here

test/cases/nvidia-training/main_test.go
```diff
@@ -31,26 +32,54 @@ func TestMain(m *testing.M) {
...
+		"INSTANCE_TYPE":   *nodeType,
+		"REGION":          region,
+		"TEAM_IDENTIFIER": *teamIdentifier,
+		"TEST_NAME":       "nvidia.training",
+	}
+
+	renderedCloudWatchAgentManifest, err := fwext.RenderManifests(manifests.CloudWatchAgentManifest, templateData)
+	if err != nil {
+		log.Fatalf("failed to render CloudWatch Agent manifest: %v", err)
+	}
```
mselim00: can you create a function to do this in `manifests` for retrieving the rendered manifests for more re-usability

test/cases/nvidia-training/main_test.go
```diff
@@ -156,6 +185,20 @@ func checkNodeTypes(ctx context.Context, config *envconf.Config) (context.Contex
...
+func getRegionFromNodes(ctx context.Context, config *envconf.Config) (string, error) {
+	clientset, err := kubernetes.NewForConfig(config.Client().RESTConfig())
+	if err != nil {
+		return "", fmt.Errorf("failed to create Kubernetes client: %w", err)
+	}
+	nodes, err := clientset.CoreV1().Nodes().List(ctx, metav1.ListOptions{Limit: 1})
+	if err != nil {
+		return "", fmt.Errorf("failed to list nodes: %w", err)
+	}
+	return nodes.Items[0].Labels["topology.kubernetes.io/region"], nil
```
mselim00: need to check that len(items) > 0 first or this risks a panic

test/manifests/assets/cloudwatch-agent.yaml
```diff
@@ -0,0 +1,103 @@
...
+            - dcgm-exporter.kube-system.svc.cluster.local:9400
+        metrics_path: /metrics
+        metric_relabel_configs:
+          - {action: replace, target_label: KubernetesVersion, replacement: "{{.VERSION}}"}
+          - {action: replace, target_label: Variant, replacement: "{{.VARIANT}}"}
+          - {action: replace, target_label: InstanceType, replacement: "{{.INSTANCE_TYPE}}"}
+          - {action: replace, target_label: TeamIdentifier, replacement: "{{.TEAM_IDENTIFIER}}"}
+          - {action: replace, target_label: TestType, replacement: "{{.TEST_NAME}}"}
+---
+apiVersion: apps/v1
```
mselim00, line 60: curious how some of these values were set b/c some of them differ from the guide in 
https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/Container-Insights-setup-metrics.html

particularly things like the requests

test/cases/nvidia-training/vars.go
mselim00: Discussed more offline, but to summarize, we should try to make this more generic. Rather than specific flags like `teamIdentifer`, we should consider a more generic one-off flag like `--metric-dimensions` that allows end users to set the desired dimensions they want for their use case. Letting the tests append one reserved dimension like `TEST_NAME` seems fine though since that's generally reusable. We also need to be careful with letting the tester assume certain set up was done in the deployer because there's no guarantee of a specific deployer being used to run the tester, so assumptions should be limited and carefully handled

test/manifests/assets/cloudwatch-agent.yaml
annapoornanarayan: Correct me if I am wrong about which values but are you referring to but for [this]( https://github.com/aws/aws-k8s-tester/pull/649/files#diff-1d92a22eaaf79683f108eebbc08d6196a5aaa362c263a7cb29a704888a8be27cR83:~:text=limits%3A,memory%3A%20200Mi ) I referred to our EKSNodeMonitoringAgent cw-agent config [here](https://code.amazon.com/packages/EKSNodeMonitoringAgent/blobs/mainline/--/e2e/monitoring/cloudwatch-agent.yaml ):

annapoornanarayan:
Thanks for the comments! Made the following changes:

- TO-DO comment for using kubeconfig for region
- Separate function for rendering CW manifest -> will be useful for nvidia and nvidia-inference tests as they follow a very similar setup
- Added a check for nodes existing before extracting labels
- Better log statement descriptions

annapoornanarayan:
- Consolidated all CW related dimension flags into `--metricDimensions` flag which takes in key value pairs. This flag remains in the test phase 
Example:
 ``` 
./nvidia-training.test \
    --test.timeout=60m \
    --test.v \
    --test.run=TestBertTraining \
    --efaEnabled=true \
    --bertTrainingImage=031349313576.dkr.ecr.us-west-2.amazonaws.com/nvidia-test:latest \
    --nodeType=p4d.24xlarge \
    --metricDimensions=TeamIdentifier=node.runtime,kubernetesVersion=1.32,amiVariant=al2023,instanceType=p4d.24xlarge,testType=nvidia.training 
```
- Unlike before there is no validation/hardcoding for these values. They are entirely upto the user and will appear as the Dimensions under `ContainerInsights/Prometheus`
- Moved the Region scraping function, CloudWatch rendering function and `--metricDimension` parsing function to the `raw.go` file so it can be easily accessed by other tests.
- Used string for parsing metric dimension because using a map involves implementing the `flag.Value` interface, which felt excessive for only one flag
- Used HTML to prevent parsing HTML chars as client.go uses `html/template` in RenderManifests. (Using raw strings leads to such strings in output`: &#34;TeamIdentifier&#34;`)
- Used this `fmt.Sprintf("          - {action: replace, target_label: %s, replacement: '%s'}", key, value)) ` to populate. Due to multiline requirement not sure how else to deal with the indentation necessity but open to suggestions.
- Modified Cloudwatch Agent YAML to allow for dynamic dimensions to be set during runtime

File: test/cases/nvidia-training/main_test.go [hunk 14]
```diff
@@ -35,22 +35,36 @@ func TestMain(m *testing.M) {
 	defer cancel()
 	testenv = env.NewWithConfig(cfg).WithContext(ctx)
 
-	manifests := [][]byte{
+	metricDimensionsMap := manifests.ParseMetricDimensions(*metricDimensions)
+	region, err := manifests.GetRegionFromNodes(ctx, cfg)
+	if err != nil || region == "" {
+		log.Printf("Warning: failed to get region from nodes. The test metrics will be emitted to us-west-2 by default: %v", err)
+		region = "us-west-2"
+	}
```
mselim00: I was just digging into this a bit more, I think for what we're doing here the simpler path is to just not specify the region. the configurion here is an override, if it's not set the region of the ec2 instance is used by default. see the agent configuration section of https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch-Agent-Configuration-File-Details.html for reference

it might make sense to take the region to publish metrics in as a flag though if we want to configure this

test/manifests/raw.go
```diff
@@ -20,4 +28,81 @@ var (
...
+		pairs := strings.Split(metricDimensions, ",")
+		for _, pair := range pairs {
+			parts := strings.SplitN(pair, "=", 2)
+			if len(parts) == 2 {
+				result[strings.TrimSpace(parts[0])] = strings.TrimSpace(parts[1])
+			}
+		}
+	}
+	return result
+}
```
mselim00: It might be worth using a flag parser that can handle a map or at least slice structure for this. Something like [gpflag](https://github.com/urfave/sflags), which we're [already using in the eksapi deployer](https://github.com/aws/aws-k8s-tester/blob/6e6dd5d69fc0d98ef2acfa2b575a2145b5d75dec/internal/deployers/eksapi/deployer.go#L105)

test/manifests/raw.go
```diff
@@ -20,4 +28,81 @@ var (
...
+		dimensions := make([]string, 0, len(metricDimensions))
+		relabelConfigs := make([]string, 0, len(metricDimensions))
+
+		for key, value := range metricDimensions {
+			templateData[strings.ToUpper(key)] = value
+			dimensions = append(dimensions, fmt.Sprintf(`"%s"`, key))
+			relabelConfigs = append(relabelConfigs,
+				fmt.Sprintf("          - {action: replace, target_label: %s, replacement: '%s'}",
+					key, value))
+		}
```
mselim00: for this you can just store it as a map in the template struct and then [format it in the template](https://pkg.go.dev/text/template). Here's a similar example, but this one was using a nested struct https://github.com/aws/aws-k8s-tester/blob/6e6dd5d69fc0d98ef2acfa2b575a2145b5d75dec/internal/deployers/eksapi/templates/unmanaged-nodegroup.yaml.template#L131

test/manifests/assets/cloudwatch-agent.yaml
mselim00: ah ok I see

annapoornanarayan:
Summary of changes: 

- Added common init() for metricDimension flag in `common/config.go`. Used plag package implementation to directly use a map in place of string parsing:
Used `gpflag` with bridging to `flag.CommandLine` for e2e framework compatibility. The e2e framework's `envconf.NewFromFlags()`requires standard Go flags, while `gpflag/pflag` create separate flag sets. Bridging resolves this incompatibility.
- From above, defined MetricDimension as `map[string]string` type
- Enhanced RenderCloudWatchManifest with proper dimension headers (quotes/commas) and simplified templating for relabeling section + removed region requirements
- Refactored global variables in vars.go to struct format (required for `gpflag` initialization)
   Updated variable naming in `main_test.go` and `bert_training_test.go`

test/cases/common/config.go
```diff
@@ -0,0 +1,24 @@
+//go:build e2e
+
+package common
```
wwvela: instead of config.go, change it to flags.go is more easily to understand

test/manifests/raw.go
```diff
@@ -20,4 +33,39 @@ var (
...
+		err := wait.For(
+			fwext.NewConditionExtension(config.Client().Resources()).DaemonSetReady(&daemonset),
+			wait.WithTimeout(5*time.Minute),
+		)
+		if err != nil {
+			return ctx, fmt.Errorf("%s daemonset is not ready: %w", name, err)
+		}
+		log.Printf("%s daemonset is ready.", name)
+		return ctx, nil
+	}
```
wwvela: These method should also move to common folder in a file like helper.go

test/cases/common/config.go
```diff
@@ -0,0 +1,24 @@
+//go:build e2e
```
mselim00: test utils like this fit more cleanly into `internal/e2e`, `test/cases` is intended for test executables

test/cases/nvidia-training/main_test.go
```diff
@@ -35,22 +36,30 @@ func TestMain(m *testing.M) {
 	defer cancel()
 	testenv = env.NewWithConfig(cfg).WithContext(ctx)
 
-	manifests := [][]byte{
+	// Render CloudWatch Agent manifest with dynamic dimensions
+	renderedCloudWatchAgentManifest, err := manifests.RenderCloudWatchAgentManifest(common.MetricDimensions)
+	if err != nil {
+		log.Printf("Warning: failed to render CloudWatch Agent manifest: %v", err)
+	}
```
mselim00: IIUC we don't want to make these requirements, so I just want to sanity check that neither this nor the cloud watch agent daemonset deployment calls are blocking

so 2 qq's:
1. would the CW agent daemonset still go ready even when the infra from the deployer isn't deployed? 
2. can the render function only error if metric dimensions is set and there's some sort of bad value in there?

test/manifests/raw.go
```diff
@@ -20,4 +33,39 @@ var (
...
+// DeployDaemonSet returns a function to deploy and wait for a DaemonSet to be ready
+func DeployDaemonSet(name, namespace string) env.Func {
+	return func(ctx context.Context, config *envconf.Config) (context.Context, error) {
+		log.Printf("Waiting for %s daemonset to be ready.", name)
+		daemonset := appsv1.DaemonSet{
+			ObjectMeta: metav1.ObjectMeta{Name: name, Namespace: namespace},
+		}
+		err := wait.For(
+			fwext.NewConditionExtension(config.Client().Resources()).DaemonSetReady(&daemonset),
+			wait.WithTimeout(5*time.Minute),
```
ndbaker1: can you also include `WithContext`?

test/manifests/raw.go
ndbaker1: +1, lets keeps anything that's not about manifest/asset manipulation in a different place

test/manifests/assets/dcgm-exporter.yaml
```diff
@@ -57,6 +59,7 @@ metadata:
     app.kubernetes.io/name: "dcgm-exporter"
     app.kubernetes.io/version: "4.1.3"
 spec:
+  clusterIP: "None"
```
ndbaker1, line 62: so you dont need the IP but you want this for like endpoint discovery?

test/manifests/assets/dcgm-exporter.yaml
```diff
@@ -26,6 +26,8 @@ spec:
         env:
         - name: "DCGM_EXPORTER_LISTEN"
           value: ":9400"
+        - name: "DCGM_EXPORTER_INTERVAL"
+          value: "100"
```
ndbaker1, line 30: was this value too high or low?

test/cases/common/config.go
```diff
@@ -0,0 +1,24 @@
...
+var MetricDimensions = make(map[string]string)
+
+func InitFlags(config interface{}) {
+	flags, err := gpflag.Parse(config)
+	if err != nil {
+		panic(err)
+	}
+	flags.StringToStringVar(&MetricDimensions, "metricDimensions", nil, "Metric dimensions as key=value pairs")
+
+	flags.VisitAll(func(pf *pflag.Flag) {
```
ndbaker1: i like struct flags, but i don't think we need this wrapper

could you maybe just make a `MetricOpts` struct that's meant to be embedded into the test that uses it? `MetricDimensions` doesn't actually need to be treated like a global does it?

the config you have below would become something like this, but ill let you work out the details
```go
type Config struct {
    MetricOpts
	BertTrainingImage string `flag:"bertTrainingImage" desc:"Docker image used for BERT training workload"`
	EfaEnabled        bool   `flag:"efaEnabled" desc:"Enable Elastic Fabric Adapter (EFA)"`
	NodeType          string `flag:"nodeType" desc:"Instance type for cluster nodes"`
}
```

test/cases/nvidia-training/vars.go
```diff
@@ -3,25 +3,26 @@
...
+}
+
 // Shared global variables
 var (
-	testenv           env.Environment
-	bertTrainingImage *string
-	efaEnabled        *bool
-	nodeType          *string
+	testenv    env.Environment
+	testConfig = &Config{}
```
ndbaker1: does this need to be a pointer? i'd prefer the zero value equivalent
```suggestion
	testConfig Config
```

test/manifests/assets/dcgm-exporter.yaml
annapoornanarayan: yes exactly! Setting clusterIP:None lets us access the pod IPs directly

test/manifests/assets/dcgm-exporter.yaml
annapoornanarayan: It was initially not set at all. We defined the value as the default rate was too slow (high).

test/manifests/raw.go
annapoornanarayan: Will add it!

test/cases/nvidia-training/main_test.go
annapoornanarayan: 1. If infra is not deployed, the CW Agent daemonset will fail as it won't find the service account cwagent. Just tested this and it will block the test. I will add a case to make sure it just gives a warning instead
2. Invalid metric dimension values will not cause the render function to fail. The function only performs basic string operations (adding quotes and commas) to format dimension keys. Invalid dimensions may result in malformed output (ex:  [[]])but this still produces valid YAML that creates a functional ConfigMap. Pods will run normally but may display unexepcted or missing dimensions in the CloudWatch dashboard but this does not cause pod failures or affect test execution.

test/cases/common/config.go
annapoornanarayan: I made the following change in the latest commit, let me know if it follows what you were expecting: I defined the MetricDimensions field as `map[string]string` in the test struct to store the parsed values, then manually registered the flag in the parsing function using `pflag.StringToStringVar()` (because gpflag cannot auto-generate flags for map[string]string types from struct tags). So the flag is now test specific but gets parsed and populated in flags without being a global variable.

annapoornanarayan:
Summary

- Renamed `config.go `to` flags.go` and moved to` internal/e2e`
- Moved manifest helpers to `common/helper.go `
- Added case in deploydaemonset to not throw error for optional dameonsets (i.e. dcgm and cloudwatch manifests)
- Added `WithContext` to error logs and changed Config var to zero value equivalent
- Moved MetricDimensions from a  global variable into the Config struct to group test flags, then updated `InitFlags()` to accept a reference to the struct field to populate when parsing

test/cases/nvidia-training/main_test.go
```diff
@@ -26,6 +27,7 @@ import (
 )
 
 func TestMain(m *testing.M) {
+	fwext.InitFlags(&testConfig, &testConfig.MetricDimensions)
```
wwvela: This looks weird. I don't think this is what Nick recommended. You can create a new struct in flag.go like 
```
type MetricOps struct {
        MetricDimensions  map[string]string `flag:"metricDimensions" desc:"CloudWatch metric dimensions as comma-separated key=value pairs"`
}
```
And embed it into Config struct
```
type Config struct {
	MetricOps
	BertTrainingImage string            `flag:"bertTrainingImage" desc:"Docker image used for BERT training workload"`
	EfaEnabled        bool              `flag:"efaEnabled" desc:"Enable Elastic Fabric Adapter (EFA)"`
	NodeType          string            `flag:"nodeType" desc:"Instance type for cluster nodes"`
}
```
For the InitFlags you just need to pass testConfig. And inside of the InitFlags method, it can be something like below. So it can compatible with both config with and without MetricOps embeded. 
```
func InitFlags(config interface{}) error {
    flags, err := gpflag.Parse(config)
    if err != nil {
        return fmt.Errorf("failed to parse flags: %w", err)
    }

    # check if config has MetricOps embeded and init if needed
    InitMetricOps()

    flags.VisitAll(func(pf *pflag.Flag) {
        flag.CommandLine.Var(pf.Value, pf.Name, pf.Usage)
    })

    return nil
}
```
You can simply check if a config has metricOps struct embed [using reflect something like this](https://go.dev/play/p/DAUU0uLz4vY).

ndbaker1: exactly yea, the idea was re-usable flags enabled by embedding the annotated struct (eg. `MetricOpts`) into a per-test struct definition.

can you also turn `InitFlags` into something more like a drop-in replacement for `gpflag.Parse` that also handles maps rather than a whole wrapper? i.e. return a `(flagSet, error)` and maybe rename to `ParseFlags`

test/common/helper.go
```diff
@@ -0,0 +1,57 @@
...
+			ObjectMeta: metav1.ObjectMeta{Name: name, Namespace: namespace},
+		}
+		err := wait.For(
+			fwext.NewConditionExtension(config.Client().Resources()).DaemonSetReady(&daemonset),
+			wait.WithTimeout(5*time.Minute),
+			wait.WithContext(ctx),
+		)
+		if err != nil {
+			// Do not fail test for optional daemonset failures
+			if name == "cwagent" || name == "dcgm-exporter" {
```
ndbaker1: can you pass an `optional` flag parameter to this function rather than hard coding which are optional?

test/common/helper.go
```diff
@@ -0,0 +1,57 @@
...
+	fwext "github.com/aws/aws-k8s-tester/internal/e2e"
+	appsv1 "k8s.io/api/apps/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/e2e-framework/klient/wait"
+	"sigs.k8s.io/e2e-framework/pkg/env"
+	"sigs.k8s.io/e2e-framework/pkg/envconf"
+)
+
+// RenderCloudWatchAgentManifest renders the CloudWatch Agent manifest with dynamic dimensions
+func RenderCloudWatchAgentManifest(cloudWatchAgentManifest []byte, metricDimensions map[string]string) ([]byte, error) {
```
ndbaker1: the original comment about moving this out of the `manifest` package was only meant for `DeployDaemonSet`.

you can move this back to `manifest` and i'd recommend the following:
1. move this func to a `rendered.go` file inside `manifests`
2. rename the embedded manifest to template and make it private (eg. `CloudWatchAgentManifest` -> `cloudWatchAgentManifestTemplate`) since only the rendered version is meant for consumption.

test/common/helper.go
ndbaker1, line 1: nit: is there a more descriptive name for the helpers we want to put in this file?  maybe `resources.go` to mean k8s resource deployments or `environment.go` for `env.Func` helpers?

annapoornanarayan: Do we like resources.go?

test/cases/nvidia-training/main_test.go
```diff
@@ -74,22 +87,22 @@ func TestMain(m *testing.M) {
...
 
 		// Wait for DaemonSets using helper
-		deployDaemonSet("nvidia-device-plugin-daemonset", "kube-system"),
-		deployDaemonSet("aws-efa-k8s-device-plugin-daemonset", "kube-system"),
-		deployDaemonSet("dcgm-exporter", "kube-system"),
-
+		common.DeployDaemonSet("nvidia-device-plugin-daemonset", "kube-system", false),
+		common.DeployDaemonSet("aws-efa-k8s-device-plugin-daemonset", "kube-system", false),
+		common.DeployDaemonSet("dcgm-exporter", "kube-system", true),
+		common.DeployDaemonSet("cwagent", "amazon-cloudwatch", true),
```
wwvela: * It is weird to deploy the dgcm-exporter and cwagent here by default and use optional bool to just avoid test failed. I think what Nick mean is to use a optional flag for the test not adding a param for the methods.
* You can use the metric dimension flag as a check directly to see if we need to deploy these 2 daemonset. Deploy them by default and let it crashing there is not good and might introduce more issues in tests.

internal/e2e/flags.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+	"flag"
+	"fmt"
+	"github.com/octago/sflags/gen/gpflag"
+	"github.com/spf13/pflag"
+	"reflect"
+)
+
+// For CloudWatch metric dimension flag
+type MetricOps struct {
```
wwvela: This flag will only applied in test phase for different tests, not deployer phase. It doesn't make sense to put it under interal parts. The whole file should go under `test/common`

internal/e2e/flags.go
```diff
@@ -0,0 +1,36 @@
...
+func ParseFlags(config interface{}) (*pflag.FlagSet, error) {
+	flags, err := gpflag.Parse(config)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse flags: %w", err)
+	}
+
+	// Handle MetricDimensions map that gpflag doesn't support
+	if _, hasField := reflect.TypeOf(config).Elem().FieldByName("MetricDimensions"); hasField {
+		field := reflect.ValueOf(config).Elem().FieldByName("MetricDimensions")
+		metricDims := field.Addr().Interface().(*map[string]string)
```
wwvela: These 2 can be in one.

internal/e2e/flags.go
annapoornanarayan: I moved it as per [this](https://github.com/aws/aws-k8s-tester/pull/649#discussion_r2240834077) comment.

test/cases/nvidia-training/main_test.go
```diff
@@ -73,23 +88,28 @@ func TestMain(m *testing.M) {
...
 		checkNodeTypes, // Dynamically check node types and capacities after device plugins are ready
 	)
 
+	if len(testConfig.MetricDimensions) > 0 {
+		testenv.Setup(
+			common.DeployDaemonSet("dcgm-exporter", "kube-system"),
+			common.DeployDaemonSet("cwagent", "amazon-cloudwatch"),
+		)
+	}
+
```
wwvela: This should move inside test.Setup() and you can wrap it with a func and comment it like deploy optional daemonset

PR #662, opened 2025-08-01
Renaming CloudWatch Stack Name

annapoornanarayan:
*Issue #, if available:*

*Description of changes:*
Change CloudWatch stack name from `'cloudwatch-{UUID}' `to `'kubetest2-eksapi-{UUID}-cw'` so janitor can find and clean it up

### DIFF
--- a/internal/deployers/eksapi/infra.go
+++ b/internal/deployers/eksapi/infra.go
@@ -479,7 +479,7 @@ func (m *InfrastructureManager) getAZsWithCapacity(opts *deployerOptions) ([]str
 
 func getCloudWatchStackName(resourceID string) (string, string) {
 	clusterUUID := strings.TrimPrefix(resourceID, ResourcePrefix+"-")
-	return fmt.Sprintf("cloudwatch-%s", clusterUUID), clusterUUID
+	return fmt.Sprintf("%s-cw", resourceID), clusterUUID
 }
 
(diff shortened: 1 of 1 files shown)


internal/deployers/eksapi/infra.go
```diff
@@ -479,7 +479,7 @@ func (m *InfrastructureManager) getAZsWithCapacity(opts *deployerOptions) ([]str
 
 func getCloudWatchStackName(resourceID string) (string, string) {
 	clusterUUID := strings.TrimPrefix(resourceID, ResourcePrefix+"-")
-	return fmt.Sprintf("cloudwatch-%s", clusterUUID), clusterUUID
+	return fmt.Sprintf("%s-cw", resourceID), clusterUUID
```
mselim00, line 482: ~~we want to retain the resource prefix here, that's `kubetest2-eksapi` ([ref](https://github.com/aws/aws-k8s-tester/blob/c2a480abe9af027d96561bca77ff45007dfde76e/internal/deployers/eksapi/deployer.go#L28)), so we can just add the cloudwatch suffix to that instead of trimming it out~~
ignore this, misread it. we could expand `cw` to be the full `cloudwatch` name for better clarity though, stack names have a 128 char limit so it should be fine

mselim00 (approved):

PR #676, opened 2025-08-16
fix(nvidia): make nvidia unit tests bubble error

ndbaker1:
*Issue #, if available:*

*Description of changes:*

The failures in the unit test are not being bubbled up from the entrypoint without `errexit`

### DIFF
--- a/test/images/nvidia/gpu_unit_tests/unit_test
+++ b/test/images/nvidia/gpu_unit_tests/unit_test
@@ -1,5 +1,9 @@
 #!/usr/bin/env bash
 
+set -o errexit
+set -o nounset
+set -o pipefail
+
 TRACE_LOG=trace.log
 TEST_TIMEOUT=1800
 BASH="/usr/bin/bash"


test/images/nvidia/gpu_unit_tests/unit_test
```diff
@@ -1,5 +1,9 @@
 #!/usr/bin/env bash
 
+set -o errexit
+set -o nounset
+set -o pipefail
```
mselim00, line 5: wouldn't this cause it to fail fast on first failure, rather than the whole suite? I thought suite-level failure was already handled with an explicit `exit` at the end if at least one test fails 
https://github.com/aws/aws-k8s-tester/blob/52246bf8ff4f543ccef7cb0b4faadaa9c926004d/test/images/nvidia/gpu_unit_tests/bash_unit#L595-L622

ndbaker1: it is yea, it handles the intermediates tests with a subshell and then exits at the end https://github.com/aws/aws-k8s-tester/blob/52246bf8ff4f543ccef7cb0b4faadaa9c926004d/test/images/nvidia/gpu_unit_tests/bash_unit#L622

`bash_unit` handles its errors correctly, we're just not reacting to it here in `unit_test` entrypoint

PR #682, opened 2025-09-02
increase ephemeral storage on the auto mode test nodes

tzneal:
*Issue #, if available:*

*Description of changes:*

Some of the K8s conformance tests need more storage. This is currently satisifed by launching instances with local storage, but that requires new instances to be launched. We can just add more storage to the nodes to avoid that.

### DIFF
--- a/internal/deployers/eksapi/node.go
+++ b/internal/deployers/eksapi/node.go
@@ -22,6 +22,7 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/klog/v2"
 	"k8s.io/utils/pointer"
 	karpv1 "sigs.k8s.io/karpenter/pkg/apis/v1"
@@ -62,6 +63,7 @@ var (
 		ekstypes.AMITypesAl2023X8664Standard: defaultInstanceTypes_x86_64,
 		ekstypes.AMITypesAl2023Arm64Standard: defaultInstanceTypes_arm64,
 	}
+	nodeClassResource = schema.GroupVersionResource{Group: "eks.amazonaws.com", Version: "v1", Resource: "nodeclasses"}
 )
 
 type nodeManager struct {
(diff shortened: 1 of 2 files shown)


internal/deployers/eksapi/node.go
```diff
@@ -150,6 +156,43 @@ func (m *nodeManager) resolveInstanceTypes(opts *deployerOptions) (err error) {
 	return nil
 }
 
+func (m *nodeManager) createNodeClass(opts *deployerOptions, k8sClient *k8sClient) error {
+	dclient, err := dynamic.NewForConfig(k8sClient.config)
+	if err != nil {
+		return fmt.Errorf("creating dynamic client, %w", err)
+	}
```
ndbaker1: thats useful, should we just put this in the [k8sClient container](https://github.com/aws/aws-k8s-tester/blob/f06b9db2cd433d61940cd107beba23b27698228d/internal/deployers/eksapi/k8s.go#L34-L38)?

internal/deployers/eksapi/node.go
```diff
@@ -150,6 +156,43 @@ func (m *nodeManager) resolveInstanceTypes(opts *deployerOptions) (err error) {
...
+	if err != nil {
+		return fmt.Errorf("getting default nodeclass, %w", err)
+	}
+	klog.Infof("got existing default nodeclass for template..")
+
+	nodeclass.SetName(m.resourceID)
+	// delete some metadata fields and the status that we don't want to duplicate
+	if metadata, ok := nodeclass.Object["metadata"].(map[string]interface{}); ok {
+		for _, k := range []string{"annotations", "labels", "creationTimestamp",
+			"finalizers", "resourceVersion", "uid", "generation"} {
```
ndbaker1: what are the leftover metadata keys?

ndbaker1: Actually I'm more just wondering if it's simpler to create an empty object and then copy/modify the spec from the reference object

internal/deployers/eksapi/node.go
tzneal: SGTM

internal/deployers/eksapi/node.go
tzneal: SGTM

PR #692, opened 2025-09-26
feat: add disruptive reboot test case

mselim00:
*Issue #, if available:*

*Description of changes:*
Adds a new test case that reboots the instance, intended to help sanity check that reboots are graceful. It's quite difficult from AWS APIs alone to determine the status of an instance after reboot (the `RebootInstance` API itself is asynchronous, EC2 status checks and SSM agent connectivity status are reconciled sparsely so they some times miss state changes). Ultimately, something must be executed on the node to determine whether or not it's running, this uses a pod to do so to keep it more as Kubernetes-oriented as possible. 

The test is based on the assumption that the pod `Exec` require kubelet responsiveness, and therefore a `143` to a command execution within a pod will decisively indicate that the node is shutting down. This is a bit of a simplification since any `SIGTERM` would lead to this state, but given the timing and the presumed clean state of the instance, it's taken to mean the reboot is starting. After this, a second pod is created, and it follows from the prior state that this pod should not start running until after the boot, since kubelet was already non-responsive or evicting existing pods.  

Sample happy path output:
```
2025/09/26 23:36:11 Starting quick test suite...
=== RUN   TestGracefulReboot
=== RUN   TestGracefulReboot/graceful-reboot
=== RUN   TestGracefulReboot/graceful-reboot/Node_gracefully_reboots
    graceful_reboot_test.go:98: Node ip-172-31-42-36.us-west-2.compute.internal corresponds to EC2 instance: i-079f0dd190956bc92
    graceful_reboot_test.go:116: Started exec into pod termination-canary-1758929771
    graceful_reboot_test.go:106: Rebooting instance i-079f0dd190956bc92 to test graceful reboot...
    graceful_reboot_test.go:113: Successfully initiated reboot of instance i-079f0dd190956bc92, waiting for pod termination-canary-1758929771 to terminate...
    graceful_reboot_test.go:123: Pod termination-canary-1758929771 was terminated
    graceful_reboot_test.go:129: Waiting up to 5 minutes for node ip-172-31-42-36.us-west-2.compute.internal to become schedulable again
    graceful_reboot_test.go:146: Node ip-172-31-42-36.us-west-2.compute.internal became ready and schedulable within 2m5.597110799s!
=== NAME  TestGracefulReboot/graceful-reboot
    graceful_reboot_test.go:153: Successfully cleaned up pod termination-canary-1758929771
    graceful_reboot_test.go:159: Successfully cleaned up pod boot-detection-1758929771
--- PASS: TestGracefulReboot (132.70s)
    --- PASS: TestGracefulReboot/graceful-reboot (132.70s)
        --- PASS: TestGracefulReboot/graceful-reboot/Node_gracefully_reboots (132.64s)
PASS
ok      github.com/aws/aws-k8s-tester/test/cases/disruptive     132.725s
```

### DIFF
--- a/test/cases/disruptive/graceful_reboot_test.go
+++ b/test/cases/disruptive/graceful_reboot_test.go
@@ -1,0 +1,161 @@
+//go:build e2e
+
+package disruptive
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"strings"
+	"testing"
+	"time"
+
+	fwext "github.com/aws/aws-k8s-tester/internal/e2e"
+
+	"github.com/aws/aws-k8s-tester/internal/awssdk"
+	"github.com/aws/aws-sdk-go-v2/service/ec2"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/util/exec"
+
+	"sigs.k8s.io/e2e-framework/klient/wait"
+	"sigs.k8s.io/e2e-framework/pkg/envconf"
+	"sigs.k8s.io/e2e-framework/pkg/features"
+)
+
+func getSleepPodTemplate(name string, targetNodeName string, duration string) corev1.Pod {
+	return corev1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: "default",
+		},
+		Spec: corev1.PodSpec{
+			Containers: []corev1.Container{
+				{
+					Name:    name,
+					Image:   "public.ecr.aws/amazonlinux/amazonlinux:2023",
+					Command: []string{"sleep", duration},
+					Resources: corev1.ResourceRequirements{
+						Requests: corev1.ResourceList{
+							corev1.ResourceCPU:    resource.MustParse("250m"),
+							corev1.ResourceMemory: resource.MustParse("64Mi"),
+						},
+						Limits: corev1.ResourceList{
+							corev1.ResourceCPU:    resource.MustParse("250m"),
+							corev1.ResourceMemory: resource.MustParse("64Mi"),
+						},
+					},
+				},
+			},
+			RestartPolicy: corev1.RestartPolicyNever,
+			NodeName:      targetNodeName,
+			Resources: &corev1.ResourceRequirements{
+				// set high pod limits to make sure the pod does not get
+				// OOMKilled, and make requests equal to qualify the pod
+				// for the Guaranteed Quality of Service class
+				Requests: corev1.ResourceList{
+					corev1.ResourceCPU:    resource.MustParse("250m"),
+					corev1.ResourceMemory: resource.MustParse("64Mi"),
+				},
+				Limits: corev1.ResourceList{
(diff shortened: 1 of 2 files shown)


test/cases/disruptive/graceful_reboot_test.go
```diff
@@ -0,0 +1,186 @@
...
+			})
+
+			if err != nil {
+				t.Fatalf("Failed to reboot EC2 instance %s: %v", instanceID, err)
+			}
+			t.Logf("Successfully initiated reboot of instance %s, waiting for pod %s to terminate...", instanceID, canaryPod.Name)
+
+			// Attempt to execute a blocking command in the pod until we get a 143, which would indicate a SIGTERM.
+			// This a reliable way to check termination since it requires direction response from Kubelet
+			var execOut, execErr bytes.Buffer
```
ndbaker1, line 115: how reliable exactly? i just want to know whether this is deterministic or highly probable.

test/cases/disruptive/graceful_reboot_test.go
```diff
@@ -0,0 +1,186 @@
...
+					corev1.ResourceMemory: resource.MustParse("64Mi"),
+				},
+			},
+		},
+	}
+}
+
+func TestGracefulReboot(t *testing.T) {
+	terminationCanaryPodName := fmt.Sprintf("termination-canary-%d", time.Now().Unix())
+	bootIndicatorPodName := fmt.Sprintf("boot-detection-%d", time.Now().Unix())
```
ndbaker1, line 74: cant you just make the pod objects here without the nodename/seletors? avoids having to make them again in the teardown

test/cases/disruptive/graceful_reboot_test.go
```diff
@@ -0,0 +1,186 @@
...
+			}
+
+			var node corev1.Node
+			if err := cfg.Client().Resources().Get(ctx, targetNode.Name, "", &node); err != nil {
+				t.Fatalf("Failed to get node %s: %v", targetNode.Name, err)
+			}
+
+			if !isNodeReady(node, bootIndicatorPod.CreationTimestamp.Time) {
+				t.Fatalf("Node %s still not ready! %+v", node.Name, node)
+			}
```
ndbaker1: wait, if you're using a canary pod to get the node name, you can just use a nodeSelector for the pod spec then you wont need this check either since it wont be scheduled?

mselim00: removed the check, it was redundant b/c of the prior wait until the pod was running anyway

test/cases/disruptive/graceful_reboot_test.go
mselim00: for now I want to say highly probable, but I think deterministic given a clean environment. the main risk I think is that the node starts rebooting before the exec. I've moved the reboot into the background with a 1 second delay to make that even less likely, but we can make this part completely deterministic if we see it flake in the future

ndbaker1 (approved):

test/cases/disruptive/graceful_reboot_test.go
mselim00: go 1.25 seems to validate against non-test go routines in a test context, so I went back to doing this serially. the RebootInstances API call is async and reboots take time, so it's still very unlikely that the exec happens too late even doing this serially. can revise as needed though

PR #796, opened 2026-05-12
fix: Add retry to initial kubelet healthz check in TestGracefulReboot

shvbsle:
*Issue #, if available:*

*Description of changes:*
The TestGracefulReboot test was failing because the initial /healthz
check encountered a transient TLS error ("tls: internal error") from
the kubelet, causing an immediate test failure before the reboot was
even attempted.

This can happen when kubelet's serving certificate is being rotated
shortly after the node becomes schedulable. The scheduler considers
the node ready (and places the canary pod), but the kubelet healthz
endpoint briefly returns TLS errors.

This change replaces the single-shot healthz check with a retry loop
that polls every 5 seconds for up to 1 minute, giving the kubelet
time to stabilize its TLS state before we rely on the endpoint for
reboot detection.

### DIFF
--- a/test/cases/disruptive/graceful_reboot_test.go
+++ b/test/cases/disruptive/graceful_reboot_test.go
@@ -73,10 +73,23 @@ func TestGracefulReboot(t *testing.T) {
 
 			// Do an initial check of the /healthz endpoint reachability to ensure we can rely on it later.
 			// This might fail even if the node is healthy if, for example, the node's security group rules
-			// do not allow ingress traffic from the control plane
-			kubeletResponsive, err := fwext.KubeletIsResponsive(ctx, cfg.Client().RESTConfig(), targetNode.Name)
-			if err != nil || !kubeletResponsive {
-				t.Fatalf("Node %s is not responding to initial /healthz checks: %v", targetNode.Name, err)
+			// do not allow ingress traffic from the control plane.
+			// Retry for up to 1 minute to handle transient TLS errors during cert rotation.
+			var kubeletResponsive bool
+			var err error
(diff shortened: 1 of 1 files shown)


test/cases/disruptive/graceful_reboot_test.go
```diff
@@ -73,10 +73,23 @@ func TestGracefulReboot(t *testing.T) {
...
 			// This might fail even if the node is healthy if, for example, the node's security group rules
-			// do not allow ingress traffic from the control plane
-			kubeletResponsive, err := fwext.KubeletIsResponsive(ctx, cfg.Client().RESTConfig(), targetNode.Name)
-			if err != nil || !kubeletResponsive {
-				t.Fatalf("Node %s is not responding to initial /healthz checks: %v", targetNode.Name, err)
+			// do not allow ingress traffic from the control plane.
+			// Retry for up to 1 minute to handle transient TLS errors during cert rotation.
+			var kubeletResponsive bool
+			var err error
+			healthCheckCtx, healthCheckCancel := context.WithTimeout(ctx, 1*time.Minute)
```
mselim00: curious about the 1 minute? I'd be good going even longer than this if we think it'd help

mselim00 (approved):
retrying this makes sense, thanks

test/cases/disruptive/graceful_reboot_test.go
shvbsle: Cool I'll do 5 mins for now. In the grand scheme of things, 5 mins should be ok.

mselim00 (approved):

PR #807, opened 2026-07-20
Adding --no-asg and --capacity-block flag support for eksapi

arsbw2802:
**Summary**
Adds a `--no-asg` flag to the `eksapi` deployer, useful for scenarios where an ASG's mixed-instances policy is a limitation — notably capacity blocks, which the ASG mixed-instances policy does not support.
Also adds a `--capacity-block` flag that sets the launch template's market type to capacity-block. It implies `--capacity-reservation`, and only works together with `--no-asg`.

**Testing**
Default (ASG) test:
```
kubetest2 eksapi --kubernetes-version=1.33 --unmanaged-nodes \
    --nodes=1 --instance-types=t3.large --user-data-format=nodeadm \
    --up --down --test=exec -- echo "with-asg test passed"
```
Output: cluster + node came up, node became Ready in ~10s, test passed, exit code 0.
```
INFO waiting for nodes to be ready nodeCount=1 timeout=5m0s
INFO nodes are ready count=1 nodes=map[ip-x-x-x-x.us-west-2.compute.internal:{}]
INFO --log-bucket is empty, no logs will be gathered!
with-asg test passed
exit code: 0
```

`--no-asg` test 
```
kubetest2 eksapi --kubernetes-version=1.33 --unmanaged-nodes --no-asg \
    --nodes=1 --instance-types=t3.large --user-data-format=nodeadm \
    --up --down --test=exec -- echo "no-asg test passed"
```
Output: cluster + node came up, `verifyAMI` confirmed the launch template AMI, node became Ready immediately, test passed, exit code 0. Nodegroup stack rendered a `Node0 AWS::EC2::Instance` and no `AutoScalingGroup`, confirming the flag behaves as intended.
```
  INFO waiting for nodes to be ready nodeCount=1 timeout=5m0s
  INFO nodes are ready count=1 nodes=map[ip-x-x-x-x.us-west-2.compute.internal:{}]
  INFO --log-bucket is empty, no logs will be gathered!
  no-asg test passed
  exit code: 0
```
```
  $ aws cloudformation describe-stack-resources \
      --stack-name "kubetest2-eksapi-<run-id>-unmanaged-nodegroup" \
      --query "StackResources[].[LogicalResourceId,ResourceType]" --output table
  ---------------------------------------------------------------------
  |                      DescribeStackResources                       |
  +--------------------------------+----------------------------------+
  |  EFASecurityGroupEgress        |  AWS::EC2::SecurityGroupEgress   |
  |  EFASecurityGroupEgressAllIpv4 |  AWS::EC2::SecurityGroupEgress   |
  |  EFASecurityGroupEgressAllIpv6 |  AWS::EC2::SecurityGroupEgress   |
  |  EFASecurityGroupIngress       |  AWS::EC2::SecurityGroupIngress  |
  |  Node0                         |  AWS::EC2::Instance              |
  |  NodeInstanceProfile           |  AWS::IAM::InstanceProfile       |
  |  NodeLaunchTemplate            |  AWS::EC2::LaunchTemplate        |
  +--------------------------------+----------------------------------+
```


internal/deployers/eksapi/deployer.go
```diff
@@ -65,6 +65,7 @@ type deployerOptions struct {
 	AMIType                     string        `flag:"ami-type" desc:"AMI type for managed nodes"`
 	AutoMode                    bool          `flag:"auto-mode" desc:"Enable EKS Auto Mode"`
 	CapacityReservation         bool          `flag:"capacity-reservation" desc:"Use capacity reservation for the unmanaged nodegroup"`
+	CapacityBlock               bool          `flag:"capacity-block" desc:"Use capacity block market type for the reservation. Implies --capacity-reservation."`
```
wwvela: Could you also add info here that capacity block is only work with --no-asg cuz mixed instance policy doesn't support it?

mselim00: if that's the only limitation, shouldn't it also work when `len(InstanceTypes) == 1`?

internal/deployers/eksapi/templates/unmanaged-nodegroup.yaml.template
```diff
@@ -116,9 +125,18 @@ Resources:
...
           Fn::If:
-            - IsCapacityReservationIdSet
+            - IsOnDemandReservation
             - CapacityReservationTarget:
                 CapacityReservationId: !Ref CapacityReservationId
+            - Fn::If:
+              - IsCapacityBlock
+              - CapacityReservationTarget:
+                  CapacityReservationId: !Ref CapacityReservationId
+              - !Ref AWS::NoValue
```
mselim00, line 135: am I missing something or is it the same for `IsOnDemandReservation` or `IsCapacityBlock`? if so the nested if could be simplified to a single if evaluating whether `CapacityReservationId` is empty or not

internal/deployers/eksapi/templates/unmanaged-nodegroup.yaml.template
```diff
@@ -180,6 +199,14 @@ Resources:
...
               {{- end}}
+      {{- else }}
+      # A single instance type uses the launch template directly. A
+      # MixedInstancesPolicy cannot be used with capacity blocks, so it is only
+      # used when more than one instance type is requested.
+      LaunchTemplate:
+        LaunchTemplateId: !Ref NodeLaunchTemplate
+        Version: !GetAtt NodeLaunchTemplate.LatestVersionNumber
+      {{- end }}
       DesiredCapacity: !Ref NodeCount
```
mselim00, line 210: can this be accomplished with the `Fn::If` and `Fn::Equals` CFN intrinsic functions instead of the go templating? it's generally better to rely on CFN built-ins where possible, though this file does use some go templating to do what CFN does not

internal/deployers/eksapi/deployer.go
```diff
@@ -310,6 +311,9 @@ func (d *deployer) verifyUpFlags() error {
 	if d.TargetCapacityReservationId != "" {
 		d.CapacityReservation = true
 	}
+	if d.CapacityBlock {
+		d.CapacityReservation = true
+	}
```
mselim00, line 316: how does this flag actually work? is it expected that the user also supplies `TargetCapacityReservationId`? if not, in the current state, this flag enabling `CapacityReservation` would just lead the deployer to find any generic capacity reservation that satisfies the request instance type and attempt to use it as a capacity block, which of course will not always work. if yes, we should make that explicit by throwing an error if it's missing

more broadly though, I'm wondering if we really need to expose this as a flag at all, I don't think the difference between capacity reservations and capacity blocks is meaningful to the end user of this project. it seems like it could be an implementation detail. we already call [DescribeCapacity reservations](https://github.com/arsbw2802/aws-k8s-tester/blob/1129600a528b467571fb8579dba517b6815c4d48/internal/deployers/eksapi/node.go#L658) to find a suitable one, and that includes a `ReservationType` that would tell us whether it's a standard CR or a CB

mselim00 (commented):
thanks for the revision! I have just a few more comments on the direction of this

internal/deployers/eksapi/deployer.go
arsbw2802: Yes, we do need `TargetCapacityReservationId` to be passed in along with this flag, and yeah I agree on making that requirement explicit by throwing an error. I can add that feature.

internal/deployers/eksapi/templates/unmanaged-nodegroup.yaml.template
arsbw2802: This can't fully be accomplished with just CFN intrinsic functions, selecting MixedInstancePolicy vs Launch template if / else should be doable with CFN intrinsic functions, however I believe the Overrides list and the condition of checking whether the number of instance types is greater than 1 needs Go templating. So what I can do is a mix of both, use CFN intrinsic function wherever possible and do the rest with Go templating?

internal/deployers/eksapi/templates/unmanaged-nodegroup.yaml.template
arsbw2802: Yes I believe it is the same for both IsOnDemandReservation and IsCapacityBlock, and can be simplified further, let me look into it.

internal/deployers/eksapi/deployer.go
arsbw2802: I looked into this further and it seems like even if TargetCapacityReservationId is not required to be passed through explicitly, it is enfroced indirectly. capacity block flag requires capacity reservation flag to also be passed through. And when CapacityReservation is true, the ID is automatically discovered and populated through getCapacityReservation. So it seems like TargetCapacityReservationId is optional. And in the case that a reservation doesn't exist, there is a hard error which causes the entire setup process to fail. However, we can still go ahead and make the TargetCapacityReservationId explicit if needed.

mselim00: I think not exposing the flag and just internally setting the market type is the way to go, we should know that from the describe call

PR #810, opened 2026-09-03
Add opt-in deep DCGM diagnostic feature (level 4)

junpengdev:
Adds a standalone `dcgm-diag` feature running `dcgmi diag -r 4`, separate from the level-2 pass the unit-test feature already performs. Level 4 (`xlong`) adds `memtest` and `pulse_test` on top of level 3.

Runtime grows with GPU count, though sub-linearly: measured with DCGM 4.6.1 and driver 580, about 32 minutes on a single-GPU instance and about 80 minutes on an eight-GPU instance. Total HBM has little effect, so budget by GPU count. NVIDIA's guidance allows up to 2.25 hours on eight GPUs.

The feature is opt-in via `--dcgmDiagEnabled` and defaults to off. This is load-bearing rather than conservative: some harnesses invoke this binary without `-test.run`, passing only `--skip-features`, so any new Test function here runs by default in those pipelines. Relying on `--skip-features` to hold it back would mean a harness that forgets an entry silently inherits an hours-long GPU stress. Failing closed puts the consequence of a mistake on the side of not running.

The Job body is a shell script baked into the nvidia test image next to `unit_test`, rather than inlined in the manifest, so it can be read and linted as shell. It resolves its DCGM config relative to its own location, so it picks up the same per-GPU-model overrides the level-2 run already relies on.

Two guards are worth calling out.

The Job name is fixed, so a Job left behind by an interrupted run on a reused cluster would let the wait observe that Job's terminal state and report success without running anything. Setup therefore deletes any pre-existing Job and waits for it to go before creating a new one, then confirms the new Job exists. Both are necessary because `fwext.ApplyManifests` can return nil even when the underlying Create failed: `processObjects` discards the per-object error (`internal/e2e/client.go`). That discard is a pre-existing repo-wide issue and is left for a separate change.

The Job's `activeDeadlineSeconds` is derived from the wait timeout, and Kubernetes rejects a Job whose `activeDeadlineSeconds` is not positive. `dcgmDiagDeadlineSeconds` keeps that bound and the arithmetic together so they cannot drift, and lives in an untagged file so it can be unit tested without the e2e TestMain, which needs a live cluster.

The datacenter-gpu-manager-4-cuda12/13 packages already in that image carry the plugins level 4 needs, so no packaging change is required.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/test/cases/nvidia/dcgm_deadline.go
+++ b/test/cases/nvidia/dcgm_deadline.go
@@ -1,0 +1,40 @@
+package nvidia
+
+import (
+	"fmt"
+	"time"
+)
+
+// dcgmDiagDeadlineSlack is how far inside the wait timeout the Job's own
+// deadline sits. Bounding the Job first means a hung run surfaces as a Job
+// failure with logs attached, rather than as an opaque wait timeout.
+const dcgmDiagDeadlineSlack = time.Minute
+
+// minDcgmDiagTimeoutMinutes is the smallest wait timeout that still yields a
+// positive activeDeadlineSeconds. Kubernetes rejects a Job whose
+// activeDeadlineSeconds is not positive, so anything lower cannot produce a
+// valid Job.
+const minDcgmDiagTimeoutMinutes = 2
+
+// dcgmDiagDeadlineSeconds converts a wait timeout in minutes into the Job's
+// activeDeadlineSeconds, rejecting values that cannot render a valid Job.
+//
+// Validation and arithmetic live together deliberately: keeping them apart
+// invites a bounds check that no longer matches the expression it guards.
+//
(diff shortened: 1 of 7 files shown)


wwvela:
The image build failed with `unrecognized import path "gitlab.alpinelinux.org/alpine/go"` for some reason. Could you check it? The changes overall LGTM

junpengdev:
> The image build failed with `unrecognized import path "gitlab.alpinelinux.org/alpine/go"` for some reason. Could you check it? The changes overall LGTM

This seems a github CI issue. I triggered a No-op PR https://github.com/aws/aws-k8s-tester/pull/811 and it still has the same issue.

wwvela (approved):

PR #814, opened 2026-09-09
Adding additional nvidia test suites for supporting multi nvidia drivers

arsbw2802:
Adding test suites for following capabilities, each capability has its own test suite + binary:
1) Driver liveness
2) Fabric Manager + NVLink
3) Cuda compute
4) Persistence device nodes 
5) GSP Firmware 

### DIFF
--- a/test/cases/nvidia-driver-liveness/main_test.go
+++ b/test/cases/nvidia-driver-liveness/main_test.go
@@ -1,0 +1,127 @@
+//go:build e2e
+
+// Package nvidia_driver_liveness is an e2e test that asserts the NVIDIA
+// driver is alive and reports the versions/models expected of the AMI
+// under test. Covers three capability #1 sub-tests: nvidia-smi runs,
+// driver version matches the -expectedDriverVersion flag, GPU
+// model matches the instance type. Runs as one pod that performs all
+// three assertions in a single bash script; the pod exits non-zero on
+// any failure with a specific code (10=smi, 11=version, 12=model).
+//
+// Excludes tests already covered upstream in nvidia.test / test_sysinfo.sh:
+// test_nvidia_gpu_count, test_nvidia_gpu_unused, test_nvidia_gpu_throttled.
+package nvidia_driver_liveness
+
+import (
+	"context"
+	_ "embed"
+	"log"
+	"os"
+	"os/signal"
+	"slices"
+	"testing"
+
+	fwext "github.com/aws/aws-k8s-tester/internal/e2e"
+	"github.com/aws/aws-k8s-tester/test/common"
+	"github.com/aws/aws-k8s-tester/test/manifests"
+	"sigs.k8s.io/e2e-framework/pkg/env"
+	"sigs.k8s.io/e2e-framework/pkg/envconf"
+)
+
+//go:embed manifests/pod-driver-liveness-check.yaml
+var podDriverLivenessCheckManifest []byte
+
+// Config is the flag surface for this test binary. Field tags become
+// `-<flag>` on the command line (parsed via common.ParseFlags).
+type Config struct {
+	NvidiaTestImage       string `flag:"nvidiaTestImage" desc:"URL of the nvidia test image (built from upstream test/images/nvidia/Dockerfile) -- required"`
+	ExpectedDriverVersion string `flag:"expectedDriverVersion" desc:"expected NVIDIA driver version (e.g. 570.86.15); when empty the 1.5 assertion is skipped"`
+	NodeType              string `flag:"nodeType" desc:"EC2 instance type under qualification (e.g. p5.48xlarge) -- required for 1.8 GPU-model regex"`
+	InstallDevicePlugin   bool   `flag:"installDevicePlugin" desc:"install the NVIDIA k8s device plugin before the test and delete it after (default true)"`
+	EfaEnabled            bool   `flag:"efaEnabled" desc:"also install the aws-efa-k8s device plugin (default false)"`
+}
+
+var (
+	testenv    env.Environment
+	testConfig Config
+)
+
+func TestMain(m *testing.M) {
+	testConfig = Config{InstallDevicePlugin: true}
+
+	if _, err := common.ParseFlags(&testConfig); err != nil {
+		log.Fatalf("failed to parse flags: %v", err)
+	}
+	cfg, err := envconf.NewFromFlags()
+	if err != nil {
+		log.Fatalf("failed to initialize test environment: %v", err)
+	}
+	if testConfig.NvidiaTestImage == "" {
+		log.Fatalf("-nvidiaTestImage is required")
+	}
+	if testConfig.NodeType == "" {
+		log.Fatalf("-nodeType is required")
+	}
+
+	ctx, cancel := signal.NotifyContext(context.Background(), os.Interrupt)
+	defer cancel()
+	testenv = env.NewWithConfig(cfg).WithContext(ctx)
+
+	// Device-plugin manifests are applied in Setup and deleted (reversed)
+	// in Finish so the cluster ends the run in the state it started in.
+	var pluginManifests [][]byte
+	var setUp []env.Func
+	if testConfig.InstallDevicePlugin {
+		pluginManifests = append(pluginManifests, manifests.NvidiaDevicePluginManifest)
+		setUp = append(setUp,
+			func(ctx context.Context, config *envconf.Config) (context.Context, error) {
+				if err := fwext.ApplyManifests(config.Client().RESTConfig(), manifests.NvidiaDevicePluginManifest); err != nil {
+					return ctx, err
+				}
+				return ctx, nil
+			},
+			common.DeployDaemonSet("nvidia-device-plugin-daemonset", "kube-system"),
+		)
+	}
+	if testConfig.EfaEnabled {
+		pluginManifests = append(pluginManifests, manifests.EfaDevicePluginManifest)
+		setUp = append(setUp,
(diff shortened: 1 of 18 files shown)


test/common/nodetype.go
mselim00, line 1: we shouldn't be hardcoding these values, this can be retrieved from `ec2:DescribeInstanceTypes`

test/cases/nvidia-driver-liveness/manifests/pod-driver-liveness-check.yaml
```diff
@@ -0,0 +1,119 @@
...
+        echo "OK 1.5: every GPU reports driver version $EXPECTED_DRIVER_VERSION"
+      fi
+
+      # ------------------------------------------------------------------
+      # 1.8  test_nvidia_gpu_model -- GPU model matches instance type.
+      # Regex per instance family. Anchored so "T4" != "T4G" and
+      # "L4" != "L40S". Skip if instance type has no mapping.
+      # ------------------------------------------------------------------
+      echo "=== 1.8 GPU model matches instance type ==="
+      case "$EC2_INSTANCE_TYPE" in
```
mselim00: the EC2 DescribeInstanceTypes API contains a partial name for the GPU. if it's always an partial match for the nvidia-smi reported product name, let's use that instead

test/cases/nvidia-fabric-manager/manifests/daemonset-fabric-manager-version-check.yaml
```diff
@@ -0,0 +1,102 @@
...
+        image: public.ecr.aws/amazonlinux/amazonlinux:latest
+        command:
+        - sh
+        - -c
+        - |
+          set -eu
+          set -o pipefail
+
+          yum install -y -q awk >/dev/null 2>&1 || \
+            dnf install -y -q awk >/dev/null 2>&1 || true
```
mselim00: this requires network connectivity, we should try to make do with a utility already present. can we not makedo with just `nvidia-smi` query options?


if we need it, let's re-use the nvidia test image. per the other test cases in here, it seems to already have awk

test/cases/nvidia-fabric-manager/manifests/daemonset-fabric-manager-version-check.yaml
```diff
@@ -0,0 +1,102 @@
...
+
+          # 5.2 only meaningful on NVSwitch nodes. The kernel driver
+          # exposes /proc/driver/nvidia-nvswitch/devices/ (populated
+          # with one entry per bound switch) only when NVSwitch
+          # hardware is present -- more reliable than an EC2 instance
+          # family list, and works on any cloud / bare metal.
+          nvswitch_dir=/host/proc/driver/nvidia-nvswitch/devices
+          if [ ! -d "$nvswitch_dir" ] || [ -z "$(ls -A "$nvswitch_dir" 2>/dev/null)" ]; then
+            echo "OK: no NVSwitch devices bound on this node; 5.2 does not apply. Skipping."
+            exec tail -f /dev/null
```
mselim00: should this just be an `exit 0`?

test/common/ec2.go
```diff
@@ -0,0 +1,69 @@
...
+	if gpuInfo == nil || len(gpuInfo.Gpus) == 0 {
+		return GPUInfo{}, fmt.Errorf("%s has no GPUs (GpuInfo is empty)", instanceType)
+	}
+	// Assume homogeneous GPUs per instance -- AWS has never mixed SKUs
+	// on a single instance type, and the API groups them accordingly.
+	g := gpuInfo.Gpus[0]
+	info := GPUInfo{}
+	if g.Count != nil {
+		info.Count = int(*g.Count)
+	}
```
mselim00, line 61: this should probably error if the assumption is broken, or at least log it

test/cases/nvidia-fabric-manager/main_test.go
```diff
@@ -34,6 +34,7 @@ var podFabricNVLinkCheckManifest []byte
 type Config struct {
 	NvidiaTestImage     string `flag:"nvidiaTestImage" desc:"URL of the nvidia test image -- required for the in-container 5.3/5.4+5.5/6.8 checks"`
 	NodeType            string `flag:"nodeType" desc:"EC2 instance type under qualification -- required for 5.3/5.4+5.5/6.8 dispatch"`
+	Region              string `flag:"region" desc:"AWS region for ec2:DescribeInstanceTypes lookup (falls back to AWS_REGION env / ~/.aws/config / IMDS when empty)"`
```
mselim00, line 37: this flag isn't necessary, the project generally assumes that the tester is configured with the same region as the nodes

test/cases/nvidia-fabric-manager/manifests/daemonset-fabric-manager-version-check.yaml
```diff
@@ -20,16 +20,17 @@ spec:
...
         - -c
         - |
           set -eu
           set -o pipefail
 
-          yum install -y -q awk >/dev/null 2>&1 || \
-            dnf install -y -q awk >/dev/null 2>&1 || true
+          # Reuses the nvidia test image so we get grep / awk / bash out
+          # of the box without a runtime package install (no network
+          # egress required from the pod).
```
mselim00: this comment is not necessary, it's about an intra-PR diff. the meaning does not carry beyond this PR

test/cases/nvidia-driver-liveness/main_test.go
```diff
@@ -112,16 +113,30 @@ func TestMain(m *testing.M) {
...
+	ExpectedGPUName       string
 }
 
-func tplVars() PodManifestTplVars {
-	return PodManifestTplVars{
+// tplVars renders the manifest values, looking up ExpectedGPUName from
+// the EC2 API. A lookup failure is logged and the check is downgraded
+// to SKIP (via empty EXPECTED_GPU_NAME) rather than failing the run --
+// missing IAM permission for ec2:DescribeInstanceTypes shouldn't
+// break qualification runs that could otherwise complete.
```
mselim00, line 130: missing `ec2:DescribeInstanceTypes` permissions **should** be breaking. silently passing tests because of a misconfigured environment creates an unnecessary risk of false qualifications

test/cases/nvidia-gsp-firmware/main_test.go
```diff
@@ -29,6 +29,7 @@ var podGSPFirmwareCheckManifest []byte
 type Config struct {
 	NvidiaTestImage     string `flag:"nvidiaTestImage" desc:"URL of the nvidia test image -- required"`
 	NodeType            string `flag:"nodeType" desc:"EC2 instance type under qualification; drives GPU count so the cross-GPU consistency check (exit 43) sees every GPU on the node"`
+	Region              string `flag:"region" desc:"AWS region for ec2:DescribeInstanceTypes lookup (falls back to AWS_REGION env / ~/.aws/config / IMDS when empty)"`
```
mselim00, line 32: same as the other comment, this flag is not necessary

test/images/nvidia/Dockerfile
```diff
@@ -51,11 +51,26 @@ RUN mkdir -p /var/run/sshd \
...
+      "2_Concepts_and_Techniques/reductionMultiBlockCG:reductionMultiBlockCG" \
+      "2_Concepts_and_Techniques/shfl_scan:shfl_scan" \
+      "0_Introduction/simpleAWBarrier:simpleAWBarrier" \
+      "0_Introduction/simpleAtomicIntrinsics:simpleAtomicIntrinsics" \
+      "0_Introduction/simpleVoteIntrinsics:simpleVoteIntrinsics" \
+      "3_CUDA_Features/warpAggregatedAtomicsCG:warpAggregatedAtomicsCG" \
+      "3_CUDA_Features/globalToShmemAsyncCopy:globalToShmemAsyncCopy"; do \
+      dir="${sample%%:*}"; bin="${sample##*:}"; \
+      cd "/tmp/cuda-samples/Samples/$dir" && cmake . && make -j$(nproc) && cp "$bin" /usr/bin || exit 1; \
+    done \
```
mselim00, line 73: nit: sorting these would be nice

mselim00 (approved):
thanks! left some more comments, but none of them need to block this PR since it's purely additive. we should address them in a followup and eventually merge the added tests into the existing ones

test/cases/nvidia-fabric-manager/fabric_manager_test.go
```diff
@@ -0,0 +1,139 @@
...
+			err := e2ewait.For(
+				fwext.NewConditionExtension(cfg.Client().Resources()).DaemonSetReady(ds),
+				e2ewait.WithTimeout(3*time.Minute),
+			)
+			if err != nil {
+				t.Logf("DS not Ready: %v", err)
+				fwext.PrintDaemonSetPodLogs(t, ctx, cfg.Client().RESTConfig(), "default", "app=nvidia-fabric-manager-version-check")
+				t.Fatalf("nvidia-fabric-manager-version-check DS not Ready within 3 minutes -- see pod logs for the 5.2 failure")
+			}
+			return ctx
```
mselim00: we need to change this check, it'll currently pass if all of the pods are ever in a `Running` state, even if they fail 2 seconds later. we specifically want to see that they all exit with a zero exit code (i.e. the pods enter a `Succeeded` state)

mselim00 (commented):
sorry I spoke too soon, we need to address the DaemonSet Ready check as well as the `exec tail` in the fabric manager test before this is ready

mselim00 (approved):
thanks!

PR #817, opened 2026-09-11
Add opt-in GPU MPS feature

junpengdev:
Adds an opt-in `mps` feature validating NVIDIA Multi-Process Service: the device plugin advertises each physical GPU several times and a control daemon funnels every client through one shared CUDA context, so their kernels run at the same time rather than the driver switching between them as under `time-slicing`.

**Three assertions:** the node advertises `physical x replicas` allocatable `nvidia.com/gpu`; N pods are observed Running simultaneously on the same GPU UUID (polling for overlap matters — a Job with parallelism N over one GPU completes even with no sharing, since each pod frees the GPU as it exits); and the GPU is in `EXCLUSIVE_PROCESS`, which only the control daemon sets. The third is what separates this from the `time-slicing` test — the first two pass identically under either.

**Opt-in** via `--mpsEnabled`, requiring the default `--installDevicePlugin`. It replaces the cluster's device plugin, so on harnesses invoking `nvidia.test` without `-test.run` an unguarded run would change GPU advertisement underneath other features.

**Restoration** drains the workload, removes the daemon, resets compute mode, restores the stock plugin — in that order, so nothing still holds a CUDA context and the daemon cannot re-apply `EXCLUSIVE_PROCESS` under the reset. It runs from both `Teardown` and `t.Cleanup`, guarded, on a context independent of the test's, since an interrupted run cancels the test context before `Teardown`. The reset is a safety net: on hardware the daemon restores `DEFAULT` itself, so it normally reports "already set to DEFAULT". A DaemonSet rather than a Job because the daemon runs on every GPU node.

Replicas default to 4, matching the Bottlerocket release tests. Builds on #816, adding the one MPS-specific hook to the shared workload: a `ReportComputeMode` flag, which `time-slicing` has no use for since `DEFAULT` is also the mode an unshared GPU is in.

**Verified** on g5.8xlarge (1x A10G) via the JobRun controller: allocatable `1 -> 4`, `4 pods ran concurrently on 1 physical GPU(s)`, 1 distinct GPU UUID across 4 pods, `Exclusive_Process` from all 4, compute mode back to `Default` and stock plugin ready. `TestMps` PASS in 425s, JobRun exit 0.

### DIFF
--- a/test/cases/nvidia/manifests/daemonset-mps-compute-mode-reset.yaml
+++ b/test/cases/nvidia/manifests/daemonset-mps-compute-mode-reset.yaml
@@ -1,0 +1,78 @@
+# Restores the GPU compute mode after the MPS feature.
+#
+# The MPS control daemon sets compute mode to EXCLUSIVE_PROCESS, which is device
+# state rather than plugin configuration. On hardware the daemon restores DEFAULT as
+# it shuts down, so on the happy path this reports "already set to DEFAULT" and does
+# nothing; it is here for the case where the daemon does not exit cleanly. Left
+# applied, the node keeps rejecting more than one CUDA context per GPU, which breaks
+# every later feature on it -- time-slicing in particular, which needs DEFAULT and,
+# because requesting time-slicing on an exclusive GPU is not an error, would fail in
+# a way that looks nothing to do with MPS.
+#
+# A DaemonSet rather than a Job because the control daemon runs on every GPU node,
+# so every one of them needs resetting. A Job with completions: 1 would return a
+# single node to DEFAULT and silently leave the rest exclusive.
+#
+# Deliberately does not request nvidia.com/gpu. This runs while the device plugin
+# is being swapped back, so a pod that needed an allocation from the plugin could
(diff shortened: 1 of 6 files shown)


test/cases/nvidia/manifests/daemonset-mps-compute-mode-reset.yaml
```diff
@@ -0,0 +1,78 @@
...
+        command:
+        - bash
+        - -c
+        - |
+          set -o pipefail
+          before=$(nvidia-smi --query-gpu=index,compute_mode --format=csv,noheader | tr '\n' ';')
+          echo "compute mode before reset: $before"
+          # -c 0 is DEFAULT. Applies to every visible GPU. Harmless when the mode
+          # is already DEFAULT, which is the usual case.
+          nvidia-smi -c 0
```
arsbw2802: Cant we just check the exit status of this to check whether all the GPUs were set to compute mode instead of explicitly checking later? Because if I understand correctly, a non-zero exit status means that all GPUs were not successfully set to compute mode?

arsbw2802 (approved):

PR #819, opened 2026-09-15
Add fixes for cuda and fabric manager tests

arsbw2802:
*Description of changes:*

For the cuda-compute tests:
Updated the naming match for the "Virtualization Mode ... : VGPU" line as the exact spacing / trailing whitespace varies across nvidia-smi versions and vGPU guest drivers, so anchoring with ^...$ was too strict.

For the Fabric Manager tests:
Updated the Pre-NVL5 detection via fabric-manager's own log line in the systemd journal. We now query both persistent (/var/log/journal) and volatile (/run/log/journal) directories. This is because distributions vary in which one holds the current boot's entries, and picking only one may miss the fabric-manager unit.
Another change is that nwe grep the systemd journal files directly rather than going through journalctl. Two reasons journalctl doesn't work here:
1. journalctl -D <dir> selects records by the container's /etc/machine-id, which doesn't match either of the machine-id subdirs mounted from the host. 
Result: it sees the store but returns no records.
2. journalctl -b picks a "current boot" using the container's /proc/sys/kernel/random/boot_id, which is inherited from the node but its journal store doesn't align with what -D exposes -- so -b filters everything out. 

journal files are binary but the log line text lives inthem as plain ASCII, so `grep -a` finds "Detected Pre-NVL5 system" exactly. The phrase is emitted only by nvidia-fabricmanager-start.sh at service startup, so zero false-positive risk.


test/cases/nvidia-fabric-manager/manifests/pod-fabric-nvlink-check.yaml
```diff
@@ -65,24 +65,30 @@ spec:
...
+            hits=$(grep -ac "Detected Pre-NVL5 system" "$jf" 2>/dev/null || true)
+            if [ "$hits" -gt 0 ]; then
+              pre_nvl5=$((pre_nvl5 + hits))
+              echo "  Pre-NVL5 probe: $jf hits=$hits (running total=$pre_nvl5)"
+            fi
+          done
+        done
+        if [ "$journal_files_checked" -eq 0 ]; then
+          echo "  note: no journal files found under /host/var/log/journal or /host/run/log/journal; cannot confirm pre-NVL5"
         else
```
mselim00: not blocking: I think we're over-complicating this a bit, can't we just mount `/var/log` to the pod's `/var/log` (not `/host/var/log` and then use `journalctl` plainly like we would on the host (w/o specifying the log dir)?

mselim00: blocking: this should exit 1. as a general rule, any issues with the test should fail closed, rather than "silently" succeeding

arsbw2802: Okay, will update

mselim00 (approved):

PR #820, opened 2026-09-15
fix(nvidia-dra): fail when the DRA kubelet plugin matches no nodes

aaithal:
*Issue #, if available:* None

*Description of changes:*

`waitForNvidiaDRADriverReady` accepts a DaemonSet that was scheduled onto nothing. `DaemonSetReady` checks `NumberReady == DesiredNumberScheduled`, trivially true at `0 == 0`, so a kubelet plugin that matched no nodes reads as ready and the suite runs GPU workloads against a driver that is not there.

The plugin selects nodes by `nvidia.com/gpu.present` through `nodeAffinity`, so "matched no nodes" is the shape a missing label takes. Wait for `DesiredNumberScheduled > 0` first, name that cause in the error, and log the node count on success. Applies to every configuration.

*Testing:*

- `gofmt` clean, `go vet -tags e2e ./test/cases/nvidia-dra/` passes, suite compiles with `go test -tags e2e -c`.
- On an EKS Auto Mode 1.35 cluster with a `g7e.2xlarge` GPU node, `waitForNvidiaDRADriverReady` logs `nvidia-dra-driver daemonset is ready on 1 node(s)`. Removing `nvidia.com/gpu.present` from that node drops `DesiredNumberScheduled` to 0 — the exact state the old check reported as ready — and setup now fails after the 5 minute timeout:

  ```
  Setup failure: nvidia-dra-driver daemonset nvidia-dra-driver-gpu/nvidia-dra-driver-gpu-kubelet-plugin
  was not scheduled onto any node (check that GPU nodes carry the nvidia.com/gpu.present=true label):
  context deadline exceeded
  ```

- DaemonSet name and namespace match the deployed chart (`nvidia/nvidia-dra-driver-gpu`, 25.8.1).

Reproducing that needs a local `g7e` entry in `instanceTopologies` — `topology.go` ships only `p5` with no override, and `TestMain` calls `log.Fatalf` on an unresolved family. Left alone here to keep this focused.


mselim00 (approved):
lgtm, thanks for raising this

we might as well make this to the `DaemonSetReady` function directly, but this is definitely a step in the right direction regardless

PR #821, opened 2026-09-17
fix: GSP firmware test skip and poll for GPU advertisement

kummatty:
*Description of changes:*

**fabric-manager**:  `countGPUNodesOfType` was doing a one of poll and failing the test because it couldn't find GPU scheduled nodes. Changing it to a Wait.

**gsp-firmware**: the EKS AMI sets `NVreg_EnableGpuFirmware=0` on g4dn/g5/g5g. It need to skip this test if EnableGpuFirmware = 0.

Verified on hardware: g5.xlarge reports `EnableGpuFirmware: 0` + GSP `N/A` (skips);
p4d.24xlarge reports GSP `580.178.04` on all 8 GPUs matching the driver (asserts).


test/cases/nvidia-gsp-firmware/manifests/pod-gsp-firmware-check.yaml
```diff
@@ -45,6 +45,19 @@ spec:
...
+      params=/host/proc/driver/nvidia/params
+      if [ -r "$params" ]; then
+        gsp_enabled=$(awk -F': *' '/^EnableGpuFirmware:/ {print $2; exit}' "$params")
+        if [ "${gsp_enabled:-}" = "0" ]; then
+          echo "SKIP 1.9: GSP disabled in the driver (EnableGpuFirmware=0); no GSP firmware to assert"
+          exit 0
+        fi
+      else
+        echo "WARN 1.9: $params unreadable; asserting GSP as if enabled"
+      fi
```
mselim00, line 56:
```suggestion
      params=/host/proc/driver/nvidia/params
      gsp_enabled=$(awk -F': *' '/^EnableGpuFirmware:/ {print $2; exit}' "$params")
      if [ "${gsp_enabled:-}" = "0" ]; then
        echo "SKIP 1.9: GSP disabled in the driver (EnableGpuFirmware=0); no GSP firmware to assert"
        exit 0
      fi
```
let's do this unconditionally. if the file does not exist or is not readable, that's a test issue that should be flagged as a failure rather than presenting a false positive

mselim00 (approved):

PR #823, opened 2026-09-21
Poll for advertised GPUs before reconfiguring the device plugin

junpengdev:
Both GPU-sharing features read allocatable nvidia.com/gpu once at the start of Setup and fail immediately on zero. That was fine when they ran on a cluster that had been up for a while; it is a race now that JobRun runs them under --test=multi with a reboot first.

The reboot tester waits for the node to be Ready, and Ready does not mean the device plugin has re-registered with kubelet and kubelet has republished the resource. In that window the count is legitimately 0, and the feature fails claiming the stock plugin is not advertising when it simply has not finished starting yet -- a misleading failure, and on a path that only exists to prove the driver came back after reboot.

Polls with the same timeout the features already use for the count to change after the plugin is swapped, since it is the same lag in the other direction. Shared, so time-slicing and MPS both get it.

Not observed failing: nvidia.timeslicing became reboot-first only recently and may not have run since. Fixing ahead of it rather than waiting for a confusing failure.

*Issue #, if available:*

*Description of changes:*


test/cases/nvidia/gpu_sharing_test.go
```diff
@@ -134,6 +134,28 @@ func allocatableGPUs(ctx context.Context, cfg *envconf.Config) (int, error) {
...
+//
+// Polling rather than reading once, because these features now run after a node reboot:
+// the reboot tester waits for the node to be Ready, and Ready does not mean the device
+// plugin has re-registered with kubelet and kubelet has republished the resource. Read
+// once, the count can still be 0 in that window, and the feature would fail claiming
+// the stock plugin is not advertising when it simply has not finished starting. The
+// same lag is already polled for on the way out, after the plugin is swapped.
+func waitForAdvertisedGPUs(ctx context.Context, cfg *envconf.Config) (int, error) {
+	var got int
+	err := wait.For(func(ctx context.Context) (bool, error) {
```
arsbw2802, line 149: Within allocatableGPUs function, we have a `cfg.Client().Resources().List(ctx, &nodes)` call. I was looking into it and I think right after a reboot, that call could very likely hit a transient API-server error in which case we would want to retry. The wait looks for any non-nil error in which case it would fail after hitting the transient error instead of continuing to poll. So maybe we could add some kind of a check here? Or maybe swallow the error and log it and retry? Just a suggestion!

arsbw2802 (approved):

arsbw2802 (approved):

